PR #1, opened 2026-02-18
Add drivers to new repo

ETSells:
Parent: https://github.com/SunDevilRocketry/Flight-Computer-Firmware/pull/228


NArmistead (approved):
never enough submodules
I do like this separation now that we have a bunch of the high level modules

PR #16, opened 2026-03-29
USB: Initial R3 Driver Implementation

### LINKED ISSUE #4: USB: Initial R3 Driver Implementation
**Requester: @ETSells **

One of our first priorities on the new project is to unlock our serial capabilities for terminal communication. If we get this ironed out, we can write test apps for pretty much every other driver. That makes USB one of our first priorities.

I would like our existing driver in `driver/usb` to have conditional definition flags for the rev 3 PCB (A0010). If these are set, we should call into the USB middleware. If they are not, we should use the existing UART TX/RX for compatibility. Since it should only be one function call that's different, this shouldn't introduce too much complexity. 

Then, port the USB middleware to `SunDevilRocketry/lib`. If you need an additional interface layer between the two, you can write those functions in `driver/usb` either as a separate file or conditionally defined in `driver/usb/usb.c`. If you need an initialization function in the driver, call it `usb_init()` for internal convention alignment.

N/A in FW, hardware considered a USB-UART bridge but it takes up a bunch of space.

**Additional context**
Trying to get @MasterUser43 on board for this as he did a really good job with it in intro. Will sync as soon as possible since he won't make the early part of the kickoff meeting.

Firmware context:
The use case for USB is not performance critical. Blocking function calls are strongly preferred to avoid introducing synchronization issues and software complexity.

Hardware context:
We'll be using an STM32H733 for this project. Unlike previous SDR production projects, we won't be using a USB-UART bridge, and USB is internal. We want the board to show up as a serial port to SDEC.

**Points of Contact**
R3 Project Lead: @ETSells

MasterUser43:
## Description
This PR implements blocking-style serial driver in `driver/usb` that abstracts Native USB CDC and Legacy UART interfaces. 

Per discussion, uses `#ifdef USE_USB_CDC_FS` compiler flag to toggle between the ST USB CDC Middleware (for the Rev 3 A0010 board) and HAL UART logic (for legacy boards using a USB-to-UART bridge). A single set of API exists that calls (`usb_transmit`, `usb_receive`, etc.) without needing to know the underlying hardware. Refer to [README.md](./usb/README.md) for markdown docs.

**Key implementations & fixes:**
* Abstracted ST's USB middleware into a blocking-style API using an internal 2048-byte circular RX buffer.
* Added deadlock prevention (`VCP_Force_Unlock`) to handle physical cable hot-swaps and disconnects without hanging the MCU (in theory).

### Issue Link
Closes [USB: Initial R3 Driver Implementation](https://github.com/SunDevilRocketry/driver/issues/4)

### Testing
As noted by @ETSells in the issue, no hardware testing method exists to verify operation. The overall usb logic 'worked' fine with a HS->FS name swap and VBUS set high on a H753 board when running a basic python echo test, but not much testing is possible. So, this overall is 'untested' and 'unverified'. The legacy UART path compiles cleanly and should not affect existing boards.

### Other
* Took the audible from @ETSells and used the generic `USE_USB_CDC_FS` flag.

This driver was based on HS with Phy FS based on H733 & generated files in /lib

Might consider adding minor checks such as a `__HAL_UART_CLEAR_OREFLAG(g_huart)` to the legacy UART `usb_flush()` function for cases such as a buffer overrun prior to flush, but in general, testing is a priority.

This was also somewhat rushed, so more refined documentation may due.


ETSells:
Hey @MasterUser43, thanks for contributing! I will get this reviewed as soon as possible.

For future reference, please leave the [pull request template](https://github.com/SunDevilRocketry/driver/blob/main/.github/pull_request_template.md) intact when you create a PR as it contains a review template for us. It also includes a framework for the information we want from a contributor that is external to the code provided. I will attach it again below for my own use:

## Reviewer Checklist

### Standards
- [x] Code is testable

### Error Handling

### Memory

### Performance
- [x] Busy waiting is avoided

usb/USB_DEVICE/App/usb_device.c
ETSells: The legacy error handler may cause issues on integration. An enum `ERROR_CODE` argument is expected for the Error_Handler alias macro and the error_fail_fast error handler. The definitions for these can be found in [error_sdr.h](https://github.com/SunDevilRocketry/mod/blob/development/appa/error_sdr/error_sdr.h).

Upon further review, I'm not sure if we ever added them for USB. Consider an `ERROR_USB_INIT_ERROR` status.

usb/USB_DEVICE/App/usbd_cdc_if.c
```diff
@@ -31,7 +31,10 @@
 
 /* USER CODE BEGIN PV */
 /* Private variables ---------------------------------------------------------*/
-
+#define USB_RX_BUF_SIZE 2048
```
ETSells, line 44: Is there a TX_BUF? I recall from the .ioc file that it was configured.

usb/USB_DEVICE/App/usbd_cdc_if.c
```diff
@@ -316,6 +334,71 @@ static int8_t CDC_TransmitCplt_HS(uint8_t *Buf, uint32_t *Len, uint8_t epnum)
...
+    // if it hits the end of the array.
+    
+    uint8_t b = USB_Internal_RxBuffer[USB_Internal_RxTail]; 
+    USB_Internal_RxTail = (USB_Internal_RxTail + 1) % USB_RX_BUF_SIZE;
+    return (int16_t)b;
+}
+
+uint8_t VCP_Is_Busy(void) 
+{
+    /* In case some idiot tries to usb transmit w/o initialization */
```
ETSells: reasonable error handling but might refrain from calling devs idiots in case they hit this case lmao

usb/USB_DEVICE/App/usbd_cdc_if.c
```diff
@@ -316,6 +334,71 @@ static int8_t CDC_TransmitCplt_HS(uint8_t *Buf, uint32_t *Len, uint8_t epnum)
...
+//        __disable_irq();                      // Brief
+//        hcdc->TxState = 0;
+//        __set_PRIMASK(primask);               // Restore previous state
+    }
+
+}
+
+void VCP_Clear_RxBuffer(void)
+{
+    __disable_irq();  // Brief critical section for pointer consistency
```
ETSells: be very careful about the use of these global IRQ disablement flags, this also pauses timer wraparounds and PWM signals and such. Fine here since the critical section is like 4 asm instructions, but worth keeping in mind. When possible we prefer disabling the specific interrupt signals that could be problematic or creating an alias for non-critical interrupts like we do to guard concurrent access to `SENSOR_DATA sensor_data`.

usb/USB_DEVICE/App/usbd_cdc_if.c
```diff
@@ -316,6 +334,71 @@ static int8_t CDC_TransmitCplt_HS(uint8_t *Buf, uint32_t *Len, uint8_t epnum)
 
 /* USER CODE BEGIN PRIVATE_FUNCTIONS_IMPLEMENTATION */
 
+
```
ETSells, line 347: since these functions are not defined static, i'd prefer that you include doxygen-style comments about the return value and the function's purpose. We don't have formal coding standards created for rev 3 yet so I'm pretty lax on most stylistic things, but we do care about having the doxygen comments so we can generate documentation.

usb/USB_DEVICE/App/usbd_cdc_if.c
ETSells: typically we reserve the `//` comments for temporary comments, and permanent notes about implementation should be done with the multiline `/* */` comments.

usb/USB_DEVICE/App/usbd_cdc_if.h
```diff
@@ -110,6 +110,37 @@ uint8_t CDC_Transmit_HS(uint8_t* Buf, uint16_t Len);
 
 /* USER CODE BEGIN EXPORTED_FUNCTIONS */
 
+/**
```
ETSells, line 113: Ah, I see they're here! Can you copy these descriptions to the source file? It's probably fine and I think the tool would pick it up the same, but for internal consistency there, it'd be nice since the ST generated functions have them.

usb/README.md
ETSells, line 1: By convention, we don't do READMEs for individual drivers. If there are integration notes you'd like to include, please leave them at the top of the .c or .h file so they're generated as doxygen documentation rather than standalone here. Some of the ST HAL files do a great job of this if you'd like an example.

usb/README.md
```diff
@@ -0,0 +1,140 @@
+# Unified USB/Serial Driver Module
+
+The USB driver (`usb.c`/`usb.h`) provides an abstraction layer for serial terminal communication. It automatically detects hardware configuration via compiler flags to switch between **Native STM32 USB CDC (Virtual COM Port)** or **Legacy UART**.
+
+### Features
+- **Consistent API**: Use the same function calls regardless of whether the board uses a USB-UART bridge or internal USB.
+- **State Connection**: Automatically detects cable disconnects/reconnects and resets the internal USB state machine to prevent terminal hangs.
+- **Blocking Design**: Designed for synchronization reliability and "blocking-esk" behavior in a non-blocking middleware environment.
```
ETSells: weird that the middleware has no support for it.

usb/README.md
```diff
@@ -0,0 +1,140 @@
...
+- **Internal Buffering**: Utilizes a circular "storage tank" for USB reception to prevent data loss during terminal command entry.
```
ETSells: circular buffers my beloved and also behated

usb/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+---
+
+### Hardware Assumptions
+
+#### 1. Target Hardware
+- **Primary**: STM32H753 or STM32H733 (Rev 3 PCB / A0010).
```
ETSells: Would only define the production units this is for, which is A0010 on the H733. We do not use the H753.

usb/README.md
```diff
@@ -0,0 +1,140 @@
...
+- **Legacy**: Any STM32 H7/L4 board using a standard UART interface (e.g., A0002, L0002, L0005).
```
ETSells: Similar to the above, we do not use the L4 line. All legacy SDR boards use the STM32H750xx

usb/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+#### 2. Configuration (CubeMX / Middleware)
+To function correctly, the following hardware configurations are assumed:
+- **For Native USB (`USE_USB_CDC_FS`)**:
```
ETSells: yes we definitely want this guide in the source/header files to be safe.

usb/README.md
```diff
@@ -0,0 +1,140 @@
...
+- **Primary**: STM32H753 or STM32H733 (Rev 3 PCB / A0010).
+- **Legacy**: Any STM32 H7/L4 board using a standard UART interface (e.g., A0002, L0002, L0005).
+
+#### 2. Configuration (CubeMX / Middleware)
+To function correctly, the following hardware configurations are assumed:
+- **For Native USB (`USE_USB_CDC_FS`)**: 
+    - `USB_OTG_HS` enabled as **Internal FS Phy** (Device Only).
+    - `USB_DEVICE` Middleware set to **Communication Device Class (CDC)**.
+    - **Clock**: The USB clock must be exactly **48 MHz**.
+    - **VBUS Sensing**: Handled manually via the `USB_DETECT` pin (typically PA10).
```
ETSells: Not typically PA10, it just happens to be PA10 right now in the working version of rev3. I would remove this as the USB_DETECT pin will always be aliased to the macro when we're given the .ioc by the hardware design group.

usb/README.md
```diff
@@ -0,0 +1,140 @@
...
+    - **USB_DETECT**: Configured as a **GPIO Input**. It is highly recommended to have an **Internal Pull-Down** enabled to prevent floating signal when unplugged.
```
ETSells: heard, will try to make sure this is configured on the init side.

usb/README.md
```diff
@@ -0,0 +1,140 @@
...
+- **For Legacy (UART)**:
+    - Standard UART TX/RX pins configured.
+    - **Baud Rate**: Typically 115200.
```
ETSells: We actually prefer 921600 baud for speed.

usb/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+---
+
+### Build-Time Selection
+The driver switches modes using the `USE_USB_CDC_FS` definition (set via Makefile or Compiler Symbols). 
+
+- **Define Set**: Driver compiles for **Native USB CDC** .
+- **Define Unset**: Driver compiles for **UART** (Legacy).
```
ETSells: sick, I like this!

consider using static asserts in the header or source file to force one or the other to be defined (XOR).

usb/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+/* If USE_USB_CDC_FS is NOT defined */
+USB_STATUS usb_init(UART_HandleTypeDef* huart);
+```
+
+#### Connection Management
+```c
+void usb_poll_connection(void);
+```
+**Note:** This must be called periodically in the main loop. It monitors the physical VBUS connection (`USB_DETECT`), starting or stopping the USB device stack if the cable is hot-swapped.
```
ETSells: in the main loop? like the application loop? if so, the implementation is not acceptable. if it's called internally within the driver, this info can be kept internally and not as an integration note.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
- Project Includes                                                               
+ Standard Includes                                                               
 ------------------------------------------------------------------------------*/
-#include "main.h"
 #include "usb.h"
+#include <string.h>
+#include <stdbool.h>
 
+/* Configuration Constants */
+#define USB_FLUSH_RX_TIMEOUT_MS   100   // Timeout for flushing receive buffer
```
ETSells: oof, these timeouts are pretty high... while subsystems that use USB are not performance critical, it's still *very* rare for us to have timeouts that exceed 10 ms, and that's on rev 2 which is incredibly slow by comparison.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+#define USB_FLUSH_TX_TIMEOUT_MS   1000  // Timeout for flushing transmit buffer
+#define USB_VBUS_DISCHARGE_MS     100   // Wait time for VBUS discharge on deinit
+
+/* ========================================================================== */
```
ETSells: nit (styling): you have two different styles for these banner comments. Recommend just picking one to use throughout.

On some of the other rev3 drivers, we're using the HAL style for banner comments.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+/* ======================== NATIVE USB CDC CONFIG  ========================== */
+/* ========================================================================== */
+#ifdef USE_USB_CDC_FS
+
+#include "usb_device.h"
+#include "usbd_cdc_if.h"
+
+// Middleware handle for state control
+extern USBD_HandleTypeDef hUsbDeviceHS;
```
ETSells: where is ownership of this handle? do we need to initialize it in main?

By convention, each driver has one peripheral handle that is owned/declared/initialized in main (e.g. an I2C handle for the IMU). I'm not familiar with the USB subsystem, is this something that should be declared there?

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+ * Maps to USB_DETECT as per SDR hardware schematic.
+ */
+#define IS_USB_CONNECTED() (HAL_GPIO_ReadPin(USB_DETECT_GPIO_PORT, USB_DETECT_PIN) == GPIO_PIN_SET)
+
+/* 
+ * Helpers implemented in usbd_cdc_if.c from usbd_cdc_if.h
+ * These allow us to bypass the standard non-blocking CDC callbacks 
+ * and force a blocking-esk design.
+ */
+int16_t  VCP_Read_Byte(void);
```
ETSells: cr: please do not re-include these prototypes as they're defined in the header.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+#include "usbd_cdc_if.h"
+
+// Middleware handle for state control
+extern USBD_HandleTypeDef hUsbDeviceHS;
+
+/* 
+ * VBUS Sensing Macro. 
+ * Maps to USB_DETECT as per SDR hardware schematic.
+ */
+#define IS_USB_CONNECTED() (HAL_GPIO_ReadPin(USB_DETECT_GPIO_PORT, USB_DETECT_PIN) == GPIO_PIN_SET)
```
ETSells: you could just use usb_detect() for this rather than defining it twice.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+            // Clear dangling transmit lock in case
+            VCP_Force_Unlock();
+        }
+        previous_vbus_state = current_vbus_state;
+    }
+}
 
+USB_STATUS usb_transmit(uint8_t* data, uint16_t size, uint32_t timeout_ms) 
+{
+    if (data == NULL || size == 0 || !IS_USB_CONNECTED()) {
```
ETSells: love the precondition check, good practice

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+    }
+
+    uint32_t tickstart = HAL_GetTick();
+    
+    /* 1. Wait for any ongoing transmission to complete */
+    while (VCP_Is_Busy()) {
+        if (!IS_USB_CONNECTED() || (HAL_GetTick() - tickstart) > timeout_ms) {
+            
+            // Force 'unlock' from ST middleware to prevent 'deadlock'
+            VCP_Force_Unlock();
```
ETSells: if this forces unlock on fail, let's make sure the timeout is enough to read/write the full length of the buffer.

usb/usb.c
ETSells: can we also check to ensure size <= buffer_size

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+    if (CDC_Transmit_HS(data, size) != USBD_OK) {
+        return USB_FAIL;
+    }
+    
+    /* 3. Block until the hardware finishes pushing the data */
+    while (VCP_Is_Busy()) {
+        if (!IS_USB_CONNECTED() || (HAL_GetTick() - tickstart) > timeout_ms) {
+            
+            // Force 'unlock' from ST middleware to prevent 'deadlock'
+            VCP_Force_Unlock();
```
ETSells: preventing deadlock within your own blocking api here is really good foresight.

ETSells: while this isn't running an RTOS, this looks like it protects its resources correctly should we ever decide to move to something like that.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+    
+    return USB_OK; 
+}
 
-/*------------------------------------------------------------------------------
-Global Variables                                                                  
-------------------------------------------------------------------------------*/
+USB_STATUS usb_receive(uint8_t* data, uint16_t size, uint32_t timeout_ms) 
+{
+    if (!IS_USB_CONNECTED() || data == NULL || size == 0) {
```
ETSells: can we also add an upper bound to size?

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+    uint16_t count = 0;
 
-/*------------------------------------------------------------------------------
- Procedures 
-------------------------------------------------------------------------------*/
+    /* Poll the buffer until requested byte count is met */
+    while (count < size) {
 
+        if (!IS_USB_CONNECTED()) {
+            return USB_FAIL;
```
ETSells, line 117: i suspect this status return might cause issues in our current terminal loop structure. consider new status enum entry "USB_DISCONNECTED"?

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+}
 
-/*------------------------------------------------------------------------------
- Preprocesor Directives 
-------------------------------------------------------------------------------*/
+void usb_poll_connection(void) 
+{
+    if (hUsbDeviceHS.pClassData == NULL) return; // Not yet initialized
+
+    static uint8_t previous_vbus_state = 2;
```
ETSells: cr: magic numbers. can you make an enum for this for future maintenance?

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
 {
-/*------------------------------------------------------------------------------
- Local Variables
-------------------------------------------------------------------------------*/
-HAL_StatusTypeDef usb_status;
+    if (!IS_USB_CONNECTED()) {
+        return 0;
+    }
 
+    return (VCP_Bytes_Available() > 0) ? 1 : 0;
```
ETSells: cr: please use stdbool if you're trying to return a boolean status for readability. we also don't like this much branching on one line for the same reason.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
 
+/* Configuration Constants */
+#define USB_FLUSH_RX_TIMEOUT_MS   100   // Timeout for flushing receive buffer
+#define USB_FLUSH_TX_TIMEOUT_MS   1000  // Timeout for flushing transmit buffer
+#define USB_VBUS_DISCHARGE_MS     100   // Wait time for VBUS discharge on deinit
+
+/* ========================================================================== */
+/* ======================== NATIVE USB CDC CONFIG  ========================== */
+/* ========================================================================== */
+#ifdef USE_USB_CDC_FS
```
ETSells, line 81: cr: having each function defined twice based on this one conditional compilation macro is a bit unreadable. two options here:

1. do the conditional compilation within the body of each function a-la sensor.c (which can get ugly if we scale to another implementation).
2. move the OTG_FS implementation to a new source file and leave the old one as it was, with maybe a static assert in case someone ever tries to build it into rev3.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
-usb_detect_pinstate = 0;
+/* --- Private Variables --- */
+static UART_HandleTypeDef* g_huart = NULL;
 
+/* --- Functions --- */
 
-/*------------------------------------------------------------------------------
- API Function Implementation 
-------------------------------------------------------------------------------*/
+USB_STATUS usb_init(UART_HandleTypeDef* huart)
```
ETSells: cr (breaking change): this API does not exist on any of the legacy platforms. Because the uart handle is required for USB communication, there will be null pointer dereferences at runtime on all legacy platforms. Must fix.

Recommend removing this usb_init construct from legacy **entirely** and preserving the old method, which used a USB_HUART define in the pindefs file.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
-#endif /* #if defined( A0002_REV2 ) || defined( FLIGHT_COMPUTER_LITE ) */
-
-
-/*******************************************************************************
-* END OF FILE                                                                  * 
-*******************************************************************************/
\ No newline at end of file
+void usb_poll_connection(void){ /* Dummy Function O.O */}
+
+USB_STATUS usb_transmit(uint8_t* data, uint16_t size, uint32_t timeout_ms)
```
ETSells: cr: While uint8_t pointers are the standard at SDR for arbitrary buffers, legacy actually used a void pointer. Changing to uint8_t without fixing every invocation to cast explicitly will cause incompatible type errors on all legacy platforms. Must fix.

Image

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+    
+    HAL_StatusTypeDef status = HAL_UART_Transmit(g_huart, data, size, timeout_ms);
+    
+    if (status == HAL_OK)      return USB_OK;
+    if (status == HAL_TIMEOUT) return USB_TIMEOUT;
+    
+    return USB_FAIL;
+}
+
+USB_STATUS usb_receive(uint8_t* data, uint16_t size, uint32_t timeout_ms)
```
ETSells: cr: While uint8_t pointers are the standard at SDR for arbitrary buffers, legacy actually used a void pointer. Changing to uint8_t without fixing every invocation to cast explicitly will cause incompatible type errors on all legacy platforms. Must fix.

Image

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
-* 	    Receives bytes from the USB port                                       *
-*                                                                              *
-*******************************************************************************/
-USB_STATUS usb_receive 
-	(
-	void*    rx_data_ptr , /* Buffer to export data to        */
-	size_t   rx_data_size, /* Size of the data to be received */
-	uint32_t timeout       /* UART timeout */
-	)
+bool usb_detect(void)
```
ETSells: cr: The usb_detect API is only provided for the new implementation. This is a breaking change on legacy platforms and will cause linker errors. usb_detect must be available on the following platforms:

A0010 (Rv 1) - Flight Computer Rev. 3
A0002_REV2 - Flight Computer Rev. 2
L0002_REV5 - Liquid Engine Controller Rev. 5
L0005_REV3 - Valve Controller Rev. 3
(FC Lite backwards compatibility not required)

It must be omitted for the following legacy platforms as there is no hardware support:
A0005_REV1 and A0005_REV2 (A0005) - Ground Station (Rev 1 and 2)

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
-	}
-} /* usb_detect */
-#endif /* #if defined( A0002_REV2 ) || defined( FLIGHT_COMPUTER_LITE ) */
-
-
-/*******************************************************************************
-* END OF FILE                                                                  * 
-*******************************************************************************/
\ No newline at end of file
+void usb_poll_connection(void){ /* Dummy Function O.O */}
```
ETSells: cr: Dummy/Stub functions should not be provided.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+    }
+    
+    HAL_StatusTypeDef status = HAL_UART_Receive(g_huart, data, size, timeout_ms);
+
+    if (status == HAL_OK)      return USB_OK;
+    if (status == HAL_TIMEOUT) return USB_TIMEOUT;
+    
+    return USB_FAIL;
+}
+
```
ETSells: cr: please remove the last three functions from this file. they do not exist on legacy and no application support/testing will be provided.

their prototypes in the header should be conditionally compiled.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
-
-/*******************************************************************************
-* END OF FILE                                                                  * 
-*******************************************************************************/
\ No newline at end of file
+void usb_poll_connection(void){ /* Dummy Function O.O */}
+
+USB_STATUS usb_transmit(uint8_t* data, uint16_t size, uint32_t timeout_ms) 
+{
+    if (g_huart == NULL || data == NULL || size == 0) {
```
ETSells: cr: g_huart invariant check and all references should be deleted and/or replaced with USB_HUART macro.

usb/usb.c
```diff
@@ -39,181 +30,305 @@
...
+USB_STATUS usb_flush(bool flush_tx) 
+{
+    if (g_huart == NULL) {
+        return USB_FAIL;
+    }
+    
+    // Clear UART receive buffer (if any)
+    if (g_huart != NULL && g_huart->Instance != NULL) {
+        while (__HAL_UART_GET_FLAG(g_huart, UART_FLAG_RXNE)) {
+            volatile uint8_t dummy = g_huart->Instance->RDR; // May be called DR on STM32F series for instance
```
ETSells: this falls under the above comment about removing these functions, but for future reference we tend to prefer to avoid STM32 sub-platform specific logic in our drivers. Per our architecture, we try to use the HAL for all register/hardware level access in accordance with a modular design. Should we develop platforms for another MCU (like the G0 series planned for rev 3's modular sub-boards), we would like to be able to use the same HAL APIs without worrying about whether there are different register naming schemes.

ETSells (changes requested):
Really strong start here!

A few breaking API changes are present here, so I'm gonna have you fix those first and then re-request review when ready.

Another note: as a general rule, the firmware team believes that a proven, tested piece of code should not be modified or refactored unless there's a bunch of work to do there in the near future. Our existing USB driver has been around for years with no issues ([cb401ab1](https://github.com/SunDevilRocketry/mod/commit/cb401ab1ac9fde1dc5ae03043d1d229e6af05138) in mod is pretty close to what it is today), so we'd rather avoid introducing logical bugs by re-writing everything when we can afford not to. I would prefer that the existing logic of that driver be preserved as close to exactly as possible so that legacy doesn't start having weird behavior when this is merged.

Also, we're gonna want to integration test the modifications to legacy before this is merged. That's something I'd be happy to try to do together.

MasterUser43:
### Simplified Prev. Comment:

> **Breaking API Changes Fixed**
> 
> - `usb_transmit` and `usb_receive` signatures corrected to `void*`/`size_t`, legacy function bodies preserved from `cb401ab1`
> - `g_huart` removed, replaced with `USB_HUART` macro throughout
> - `usb_detect` guard extended to `L0002_REV5` and `L0005_REV3`
> - `usb_flush` legacy implementation added
> - `USB_DISCONNECTED` added to `USB_STATUS` for mid-receive cable loss
> - Bounds checks added to `usb_transmit` and `usb_receive`
> - Flush timeouts reduced to 10ms, extracted as named constants
> 
**Functions Removed**
 `usb_poll_connection`, `usb_deinit`, and `usb_data_available` removed per CR. Hot-swap via `USBD_Start`/`USBD_Stop` is implementable internally without exposing a public function — flagging in case that's desired.

> **`usbd_cdc_if.c`**
> 
> - Doxygen blocks added to all  VCP functions
> - `//` comments converted to `/* */` throughout
> - `__disable_irq()` usage documented; open to a more targeted IRQ disable if preferred
> 
> **`usb_device.c`**
> 
> - `Error_Handler` updated to `Error_Handler(ERROR_USB_CDC_INIT_ERROR)`, enum entry added to `error_sdr.h`
> 
 **Questions for review**
 - Is `USE_USB_UART` needed as an explicit legacy define, or is absence of `USE_USB_CDC_FS` sufficient?
 - Preference on comment style inside function bodies — banner block, `/* API implementation */`, or none?
- Preference on duplicating Doxygen from `.h` into `.c`, or header-only?

ETSells (changes requested):
Per our conversation today, you still want to split the driver to two files, so I'm gonna request changes to clear the review status and allow you to re-request when ready.

I went ahead and did a quick build, and it looks like your driver works except for a syntax error at EOF (that last line is not in a comment).

MasterUser43:
Re-requested review. Implemented the split file strat. as discussed.

`usb.c` acts as the single compilation unit and dispatches to `usb_cdc.c` or `usb_uart.c` via `#include` based on `USE_USB_CDC_FS`. Legacy boards continue to include `usb.c` as before with no build system changes required.

Considered hot-swap capability (i.e. folding `USBD_Stop`/`USBD_Start` internally into `usb_transmit`/`usb_receive` ). A bit of work revealed that it's "possible", but re-enum timing is more risky and because of that, it inherently defeats the point of hot-swap. The current `USB_DISCONNECTED` return is probably sufficient for the terminal use case. Left it out for now but happy to revisit if desired.

Still personally confused on the comments situation and general stylistic choices, but made it more consistent anyway and left the doxygen in headers only for maintainability purposes. If you insist on `.c` files copy pasting from the `.h`, that's fine, too.

I did not edit the files in a text editor this time, so hopefully no silly errors.

usb/usb_uart.c
ETSells, line 1: if its easier, you could legit just paste the old file with a different name. no need to reinvent the wheel or anything since this'll now need to pass review.

usb/usb_uart.c
```diff
@@ -0,0 +1,137 @@
...
+
+/* Transmits a specified number of bytes over USB */
+USB_STATUS usb_transmit
+	(
+	void*    tx_data_ptr  ,
+	size_t   tx_data_size ,
+	uint32_t timeout
+	)
+{   /* ==================== API implementation =================== */
+
```
ETSells, line 40: nit: first level should be unindented (global comment)

usb/usb_uart.c
```diff
@@ -0,0 +1,137 @@
...
+    /* Transmit bytes over UART */
+    usb_status = HAL_UART_Transmit( &( USB_HUART ),
+                                    tx_data_ptr   , 
+                                    tx_data_size  , 
+                                    timeout );
+
+    /* Return HAL status */
+    if ( usb_status != HAL_OK ) {
+        return usb_status;
```
ETSells: cr: HAL_StatusTypeDef does not map 1:1 with USB_STATUS, which means this will no longer report the timeout status correctly and will break existing functionality.

The old logic accounted for this. Recommend reversion.

usb/usb_uart.c
```diff
@@ -0,0 +1,137 @@
...
+            return USB_FAIL;
+            break;
+            }
+    }
+
+} /* usb_receive */
+
+
+/* Detects an active USB connection via VBUS GPIO */
+#if defined( A0002_REV2           ) || \
```
ETSells, line 111: oof i wish we made this easier on you, but good handling.

usb/usb_uart.c
```diff
@@ -0,0 +1,137 @@
...
+    defined( L0005_REV3           )
+bool usb_detect( void )
+{   /* ==================== API implementation =================== */
+
+    uint8_t usb_detect_pinstate;    /* USB detect state, return value from HAL    */
+
+    usb_detect_pinstate = 0;
+
+    /* Read voltage on USB detect pin */
+    usb_detect_pinstate = HAL_GPIO_ReadPin( USB_DETECT_GPIO_PORT, USB_DETECT_PIN );
```
ETSells: nit: HAL_GPIO_ReadPin returns an enum. Recommend explicitly checking the values:

typedef enum
{
  GPIO_PIN_RESET = 0U,
  GPIO_PIN_SET
} GPIO_PinState;

for clarity. Functionally equivalent though, so this is fine.

usb/usb_uart.c
```diff
@@ -0,0 +1,137 @@
...
+        return false;
+    } else {
+        return true;
+    }
+} /* usb_detect */
+#endif /* A0002_REV2 || FLIGHT_COMPUTER_LITE || L0002_REV5 || L0005_REV3 */
+
+
+/* Flushes pending data from the receive buffer */
+void usb_flush( void )
```
ETSells, line 128: Warning: this'll probably go unused and since it affects legacy platforms, I'll require testing before merging it. However, thanks for taking the time to do this, and I hope we'll have a good use for it eventually :)

usb/usb.h
```diff
@@ -33,65 +36,95 @@ extern "C" {
...
+#endif /* USE_USB_CDC_FS */
+
+/**
+  * @brief  Transmits a specified number of bytes over USB.
+  * @param  tx_data_ptr:  Pointer to the data buffer to send.
+  * @param  tx_data_size: Number of bytes to transmit.
+  * @param  timeout:      Maximum time (in ms) to wait before giving up.
+  * @retval USB_STATUS: USB_OK, USB_TIMEOUT, or USB_FAIL.
+  */
+USB_STATUS usb_transmit(void* tx_data_ptr, size_t tx_data_size, uint32_t timeout);
```
ETSells, line 77: so glad these APIs are platform independent! it'll allow drivers and internal command handlers to continue to use the APIs even if the implementation differs.

usb/usb.h
```diff
@@ -33,65 +36,95 @@ extern "C" {
...
+  * @brief  Initializes the hardware for serial communication.
+  * @note   For CDC mode (USE_USB_CDC_FS), starts the USB middleware stack.
+  *         For legacy boards, validates the provided UART handle.
+  * @param  huart: Pointer to the UART handle (legacy mode only).
+  * @retval USB_STATUS: USB_OK on success, USB_FAIL if handle is NULL.
+  */
+#ifdef USE_USB_CDC_FS
+USB_STATUS usb_init(void);
+#else
+USB_STATUS usb_init(UART_HandleTypeDef* huart);
```
ETSells: nit: consider removal, this is not necessary

usb/usb.h
```diff
@@ -33,65 +36,95 @@ extern "C" {
...
+#include <stddef.h>
+
+#ifndef USE_USB_CDC_FS
+    #include "main.h"   /* Required for UART_HandleTypeDef in legacy mode */
+#endif
 
 
 /*------------------------------------------------------------------------------
- Typdefs 
+ Typedefs
```
ETSells, line 47: dude this is misspelled everywhere in the firmware LOL

we'll get 'em one step at a time 🫡

usb/usb.h
```diff
@@ -33,65 +36,95 @@ extern "C" {
  Includes 
 ------------------------------------------------------------------------------*/
 #include <stdbool.h>
+#include <stddef.h>
+
+#ifndef USE_USB_CDC_FS
+    #include "main.h"   /* Required for UART_HandleTypeDef in legacy mode */
```
ETSells: no issue with including main.h everywhere if you want! unless you tried to build and there was recursive garbage, which ngl wouldn't surprise me.

ETSells: you can also include "stm32h7xx_hal.h" explicitly for this case rather than getting drive-by from main.h.

usb/usb.h
```diff
@@ -1,24 +1,27 @@
...
+  ******************************************************************************
+  * @file           : usb.h
+  * @brief          : Universal Serial Communication Driver
+  * @author         : Sun Devil Rocketry Firmware Team
+  *
+  * @note  Provides a unified interface for both Native USB CDC (A0010) and
+  *        Legacy UART-to-USB bridges via the USE_USB_CDC_FS compiler flag.
+  *
+  *        Features:
+  *        - Consistent API regardless of underlying hardware
```
ETSells, line 11: need my lfg emoji on github for this

usb/usb.c
```diff
@@ -48,172 +78,58 @@
...
-{
 /*------------------------------------------------------------------------------
- Local Variables
+ Procedures 
 ------------------------------------------------------------------------------*/
-HAL_StatusTypeDef usb_status;
 
+/* Validates UART handle for legacy UART-to-USB bridge boards */
+#ifndef USE_USB_CDC_FS
+USB_STATUS usb_init( UART_HandleTypeDef* huart )
```
ETSells: this will definitely go unused, fair warning. the uart handle is globally defined *everywhere*

usb/usb.c
```diff
@@ -48,172 +78,58 @@
...
 /*------------------------------------------------------------------------------
- Initializations 
-------------------------------------------------------------------------------*/
-usb_detect_pinstate = 0;
-
+ Build Configuration Notice
 
-/*------------------------------------------------------------------------------
- API Function Implementation 
+ Include the appropriate implementation source based on build configuration.
```
ETSells: doc: should actually work regardless since $(DRIVER_DIR)/usb will be in the include path already because of usb.h. That's why i recommended the preprocessor route.

usb/usb.c
```diff
@@ -48,172 +78,58 @@
...
-else
-	{
-	return true;
-	}
-} /* usb_detect */
-#endif /* #if defined( A0002_REV2 ) || defined( FLIGHT_COMPUTER_LITE ) */
+#ifdef USE_USB_CDC_FS
+    #include "usb_cdc.c"
+#else
+    #include "usb_uart.c"
```
ETSells, line 127: Before merging, I want to make sure this doesn't make GDB super hard to use since the "source" file is like 100% preprocessor includes lmao

ETSells (changes requested):
One CR left and then a bunch of nitpicks.

Almost there! Then I'd be happy to sit down and verify all the legacy stuff so we can merge it :)

usb/usb.c
```diff
@@ -1,44 +1,74 @@
...
+  *        For Native USB CDC (USE_USB_CDC_FS):
+  *        - USB_OTG_HS enabled as Internal FS Phy (Device Only)
+  *        - USB_DEVICE Middleware set to Communication Device Class (CDC)
+  *        - USB clock must be exactly 48 MHz
+  *        - VBUS sensing handled via USB_DETECT GPIO (configured as Input
+  *          with Internal Pull-Down to prevent floating when unplugged)
+  *
+  *        For Legacy UART:
+  *        - Standard UART TX/RX pins configured
+  *        - Baud rate: 921600
```
ETSells, line 20: doc: this is not an assumption that should be made. the uart baud is set by the integrating project and the terminal side together. We have occasionally done 115200 for stability, for example. 

However, I do think it's currently true that the baud rate is 921600 across the main supported projects, so this isn't a CR.

usb/usb.c
```diff
@@ -48,172 +78,56 @@
...
+{   /* ==================== API implementation =================== */
+
+    #ifdef USE_USB_CDC_FS
+        /* Initialize Native USB CDC Middleware */
+        MX_USB_DEVICE_Init();
+        return USB_OK;
+    #else
+        /* For legacy mode, the global USB_HUART is initialized in main.c.
+        We can verify CubeMX assigned the base address. */
+        if (USB_HUART.Instance == NULL) {
```
ETSells, line 116: yep this is fine! i can work with this since it's not required for use on the uart side but also provides API consistency :)

usb/usb_cdc.c
```diff
@@ -0,0 +1,158 @@
...
+
+/* Transmits a specified number of bytes over USB */
+USB_STATUS usb_transmit
+    (
+    void*    tx_data_ptr  ,
+    size_t   tx_data_size ,
+    uint32_t timeout
+    )
+{   /* ==================== API implementation =================== */
+
```
ETSells, line 36: nit: first level should be unindented (global comment)

usb/usb_cdc.c
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    return USB_OK;
+
+} /* usb_receive */
+
+
+/* Detects an active USB connection via VBUS GPIO */
+bool usb_detect( void )
```
ETSells, line 112: i think this implementation might be shared! if you want, you could move this back to usb.c to reduce duplication

usb/usb_uart.c
ETSells: nit: still present (unresolved)

usb/usb_uart.c
```diff
@@ -0,0 +1,141 @@
...
+
+} /* usb_receive */
+
+
+/* Detects an active USB connection via VBUS GPIO */
+#if defined( A0002_REV2           ) || \
+    defined( FLIGHT_COMPUTER_LITE ) || \
+    defined( L0002_REV5           ) || \
+    defined( L0005_REV3           )
+bool usb_detect( void )
```
ETSells, line 115: see usb_cdc.c, but since these have the same implementation, they could be moved to reduce code duplication

ETSells (approved):
A few nits and a documentation change request.

Build succeeds on legacy w/ no warnings. As always, requires testing before merge. Would be happy to sit down and do this with you (actually, I might go in early today and test it for you since I won't have time after).

Reminder: you don't have to fix every nitpick. If you choose not to fix something, please respond to the comment with "NAT: [justification]". If you do fix something, this does **NOT** require re-review before merging as long as your change is limited to fixing the comment.

Sorry this took so many rounds! Most people's first PRs with us take some time, even if it's a two line change. Considering you're new to firmware and took on a whole driver alone, there was more to cover regarding our standards and practices just due to sheer length. You did a really great job on this, well done!

ETSells:
Tests look good

ETSells:
@MasterUser43 Target branch switched to staging as this has no impact on v2.6.0.

Merging on your behalf so we don't diverge too far.

PR #28, opened 2026-05-01
IMU | LSM6DSV320X SPI DMA Driver

### LINKED ISSUE #7: IMU: Initial R3 Driver Implementation
**Requester: @ETSells **

With the IMU and barometer, we'll have the two main flight computer sensors up and running.

I would like at least blocking APIs for data retrieval from the IMU (but a non-blocking form would be preferred). Init interface *should* be blocking.

We could retrieve euler gyro readings like rev 2, but I would encourage us to use the SFLP (sensor fusion low power) interface in order to get quats.

Datasheet: https://www.st.com/resource/en/datasheet/lsm6dsv320x.pdf

***Performance Spec***: 
For our initial implementation, the target is 1 kHz (read accel and gyro data in under 1 ms). Configure an ODR of 1.92 kHz.

N/A

**Additional context**

**Points of Contact**
Add a tag for the component owners, leads, and/or a senior engineer to serve as primary point of contact.

MasterUser43:
## Description
This PR implements a unified IMU abstraction layer and the specific driver for the **LSM6DSV320X (High-G)** sensor, targeting the STM32H733 (A0010 Rev 3 PCB). 

The implementation introduces a modular architecture that separates hardware-specific translation layers (`imu_lsm.c`, `imu_legacy.c`) from the core flight software contract defined in `imu.h`. This allows individual projects to swap IMU backends without changing flight logic.

### Key Changes
*   **Unified System Interface:** Provides a consistent API (`imu_system_update`) that abstracts the differences between legacy 9-axis I2C architectures and the newer 6-axis SPI-DMA architecture.
*   **Asynchronous DMA Path:** Utilizes `HAL_SPI_TransmitReceive_DMA` for high-frequency FIFO burst reads, minimizing CPU overhead.
*   **On-Chip Sensor Fusion (SFLP):** Full integration with the ST SFLP engine, providing game rotation vectors (quaternions), gravity vectors, and gyroscope bias estimates directly from the hardware.
*   **High-G Runtime Transitions:** Implements logic for seamless switching between the Low-G (±16g) and High-G (±320g) sensing chains during flight if desired.

### Documentation & Readability
The code is heavily annotated with citations from the **DS14623 (Rev 3)** datasheet to support long-term maintenance and hardware audits. 

### Issue Link
*   Associated issue: [IMU R3 #7](https://github.com/SunDevilRocketry/driver/issues/7)

### Child PR Dependencies
*   **PR#1 (Library):** [LSM6DSV320X-PID Submodule](https://github.com/SunDevilRocketry/lib/pull/22)
*   **PR#2 (Firmware):** [DMA Linker Script Update](https://github.com/SunDevilRocketry/Flight-Computer-Firmware/pull/291)

### Testing [NOT APPLICABLE]
- [ ] Unit tests modified (link the test changes as a child PR)

### Other

### Standards
- [ ] Code is testable

### Error Handling

### Memory

### Performance
- [ ] Busy waiting is avoided

### Additional Elements
- [ ] SPI transactions are non-blocking where possible (DMA)
- [ ] D-Cache coherency handled for STM32H7


ETSells:
Ah sorry I just saw this! I've pinged myself as a reviewer and will take a look when I get a minute.

Changing your target branch to reflect our move to an unstable trunk system.

ETSells:
Also, I've added @266-750Balloons as this introduces DMA, which I'm less familiar with than CPU-bound interrupts.

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+  *         uncertainty on Cortex-M7. 64-byte ceiling is conservative; the
+  *         largest expected single write is 6 bytes (SFLP gbias init).
+  *         __NOP() guards satisfy CS setup/hold time at high SPI clock rates.
+  * @param  handle: Unused (imu_ctx.handle is NULL).
+  * @param  reg:    7-bit register address.
+  * @param  buf:    Data bytes to write.
+  * @param  len:    Number of data bytes (address byte excluded).
+  * @retval 0 on success, 1 on HAL error or oversized request.
+  */
+static int32_t platform_spi_write
```
ETSells, line 263: cr: Please use a status type in your final version (you can reuse the old IMU status enum if you want, else if this stays separate you'll need a new one.) <-- ignore; all external interfaces follow the status return convention which is good enough for me

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+if ( len >= SPI_WRITE_BUF_MAX ) {
+    return 1;
+}
+
+/* RW = 0 for write: bit[7] clear (DS14623 §5.1.3) */
+tx[0] = reg & (uint8_t)( ~IMU_SPI_READ_BIT );
+memcpy( &tx[1], buf, len );
+
+HAL_GPIO_WritePin( IMU_CS_GPIO_PORT, IMU_CS_PIN, GPIO_PIN_RESET );
+__NOP(); __NOP();
```
ETSells, line 295: Interesting that these are needed here -- we've never had to do NOPs since the procedure call overhead is usually sufficient delay between writing CS low.

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+    )
+{
+#define SPI_WRITE_BUF_MAX  ( 64U )
+
+uint8_t           tx[ SPI_WRITE_BUF_MAX ];
+HAL_StatusTypeDef hal_status;
+
+(void)handle;
+
+if ( len >= SPI_WRITE_BUF_MAX ) {
```
ETSells: nit: max implies that the upper bound is inclusive.

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+memcpy( &tx[1], buf, len );
+
+HAL_GPIO_WritePin( IMU_CS_GPIO_PORT, IMU_CS_PIN, GPIO_PIN_RESET );
+__NOP(); __NOP();
+hal_status = HAL_SPI_Transmit( &IMU_SPI, tx, (uint16_t)( len + 1U ),
+                                IMU_SPI_TIMEOUT_MS );
+__NOP(); __NOP();
+HAL_GPIO_WritePin( IMU_CS_GPIO_PORT, IMU_CS_PIN, GPIO_PIN_SET );
+
+return ( hal_status == HAL_OK ) ? 0 : 1;
```
ETSells: nit: we don't usually like inline logic like this since its harder to parse through

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+  */
+static int32_t platform_spi_write
+    (
+    void*          handle,
+    uint8_t        reg,
+    const uint8_t* buf,
+    uint16_t       len
+    )
+{
+#define SPI_WRITE_BUF_MAX  ( 64U )
```
ETSells: Can we define this scoped to the header file for visibility? Either that or I guess leave like an implNote in your doxygen header.

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+HAL_Delay( ms );
+} /* platform_delay_ms */
+
+
+/*------------------------------------------------------------------------------
+ Internal Helpers
+------------------------------------------------------------------------------*/
+
+/**
+  * @brief  Converts a 16-bit IEEE 754 half-precision float to float32.
```
ETSells: yoooo this is actually super useful. consider moving this to mod/math_sdr/math_sdr.c

awareness ping @NArmistead that we could use this when we serialize certain data points

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+  *         from the unit quaternion constraint: W = sqrt(1 − x² − y² − z²).
+  *         If numerical noise causes |xyz|² > 1, XYZ is normalised first so
+  *         the output remains a valid unit quaternion.
+  *         Source: ST community example (STMems_Standard_C_drivers issue #182).
+  * @param  quat:  Output [x, y, z, w] indexed [0..3].
+  * @param  sflp:  Three raw half-float values from the FIFO slot.
+  */
+static void sflp2q
+    (
+    float          quat[4],
```
ETSells, line 387: Note for the future: Our eventual implementation should use the same name-equivalent data structure as rev 2 (cc: @NArmistead), which will end up being likely a 4-float struct like this. This is more than sufficient for now as we haven't implemented that data structure.

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+
+/*------------------------------------------------------------------------------
+ Public API
+------------------------------------------------------------------------------*/
+
+/**
+  * @brief  Initializes the LSM6DSV320X IMU (Blocking).
+  * @note   Sequence:
+  *           1. Wire PID callbacks.
+  *           2. Disable INT1 EXTI (prevents spurious ISR during config).
```
ETSells, line 547: future note: we'll also need to disable interrupts on this SPI handle at the same time.

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+lsm6dsv320x_data_rate_t     g_odr;
+lsm6dsv320x_fifo_xl_batch_t xl_bdr;
+lsm6dsv320x_fifo_gy_batch_t g_bdr;
+
+
+/*--------------------------------------------------------------------------
+ Initializations
+--------------------------------------------------------------------------*/
+imu_sflp_enabled  = false;
+imu_dma_ready     = false;
```
ETSells: doesn't look like we have DMA r/w callbacks available but i'm glad we've started laying the groundwork for them!

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+        xl_bdr = LSM6DSV320X_XL_BATCHED_AT_3840Hz;
+        g_bdr  = LSM6DSV320X_GY_BATCHED_AT_3840Hz;
+        break;
+    case IMU_ODR_7680HZ:
+        xl_odr = LSM6DSV320X_ODR_AT_7680Hz;
+        g_odr  = LSM6DSV320X_ODR_AT_7680Hz;
+        xl_bdr = LSM6DSV320X_XL_BATCHED_AT_7680Hz;
+        g_bdr  = LSM6DSV320X_GY_BATCHED_AT_7680Hz;
+        break;
+    case IMU_ODR_1920HZ: /* fall-through to default */
```
ETSells, line 685: for defensive programming, consider using a debug assert here that ODR == 1920, that way in debug mode we catch that an improper param was passed.

the construct was just added in mod/error_sdr/error_sdr.h:debug_assert

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+        g_bdr  = LSM6DSV320X_GY_BATCHED_AT_1920Hz;
+        break;
+    }
+
+/* Apply loop-coupled ODR to accelerometer and gyroscope */
+pid_status |= lsm6dsv320x_xl_data_rate_set( &imu_ctx, xl_odr );
+pid_status |= lsm6dsv320x_gy_data_rate_set( &imu_ctx, g_odr  );
+
+/*
+ * Full-scale and ODR configuration - two separate accelerometer chains.
```
ETSells: just a heads up that in the future we'll want this to be runtime configurable! i'll probably ask you to write the config api for that when we're ready, but we'd want high-G mode during boost and switch to low-G in coast and beyond.

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+/**
+  * @brief  Mid-flight Accelerometer Full-Scale transition (Runtime).
+  * @note   Handles both same-chain FS changes and low-G ↔ high-G boundary
+  *         crossings. On crossing from low-G to high-G: writes CTRL1_XL_HG
+  *         (4Eh), forces low-G to ±16g HP mode (§6.1.2), and reroutes FIFO
+  *         batching (XL_HG_BATCH_EN on, low-G batch off). On crossing from
+  *         high-G to low-G: clears XL_HG_REGOUT_EN, restores low-G batching.
+  *         DS14623 Rev3 §6.1.2, Tables 70, 149, 150; COUNTER_BDR_REG1 (0Bh).
+  */
+IMU_STATUS imu_set_accel_fs
```
ETSells: oh dang, ignore me earlier, you already did it!!

imu/imu_t.c
```diff
@@ -0,0 +1,1414 @@
...
+ */
+SCB_CleanDCache_by_Addr( (uint32_t*)imu_dma_tx_buf,
+                          (int32_t)IMU_DMA_BUF_BYTES_ALIGNED );
+SCB_InvalidateDCache_by_Addr( (uint32_t*)imu_dma_rx_buf,
+                               (int32_t)IMU_DMA_BUF_BYTES_ALIGNED );
+
+/* Assert CS and launch DMA - CS de-asserted in imu_process_async_cb() */
+HAL_GPIO_WritePin( IMU_CS_GPIO_PORT, IMU_CS_PIN, GPIO_PIN_RESET );
+__NOP(); __NOP();
+hal_status = HAL_SPI_TransmitReceive_DMA( &IMU_SPI,
```
ETSells, line 1444: ohhhh dang! you've got the async interface too?!

ETSells (approved):
dude this is actually crazy...

lgtm honestly, ignore my CR comments below, I'd take this as-is...

ETSells:
Also -- because of the divergence between this version and the old IMU, we should consider keeping these separate with maybe a guard at the top of imu.h that points to imu_t.h if we're not using the legacy IMU. Don't worry about merging to the old platform unless you feel its the smarter design choice.

Because this is kept separate from the old IMU, we can merge as-is once the PID driver is imported somewhere (either driver or lib)

ETSells:
This is in my review queue, but it might take me a second to get there, just as a heads up.

imu/imu_legacy.c
```diff
@@ -0,0 +1,1282 @@
...
+*                                                                              
+*       This software is licensed under terms that can be found in the LICENSE 
+*       file in the root directory of this software component.                 
+*       If no LICENSE file comes with this software, it is covered under the   
+*       BSD-3-Clause.                                                          
+*                                                                              
+*       https://opensource.org/license/bsd-3-clause          
+*
+*******************************************************************************/
+#if !defined( A0010 )
```
ETSells: Consider:

```
#ifndef( A0002_REV2 )
#error "Unsupported hardware platform
#endif
```

rather than wrapping the whole thing in a preprocessor directive.

imu/imu_legacy.h
```diff
@@ -0,0 +1,523 @@
...
+    int16_t    gyro_y ;
+    int16_t    gyro_z ;
+    int16_t    mag_x;
+    int16_t    mag_y;
+    int16_t    mag_z;
+    uint16_t    mag_hall;
+} IMU_RAW;
+
+/* Processed IMU aata */
+typedef struct _STATE_ESTIMATION {
```
ETSells: note for myself: beware of conflicts

imu/imu_legacy.h
ETSells, line 1: Assumption: I'm assuming this has been copied over with no functional changes from the previous iteration. If I'm wrong, then please fix.

imu/imu_legacy.c
ETSells, line 1: Assumption: I'm assuming this has been copied over with no functional changes from the previous iteration. If I'm wrong, then please fix.

ETSells (changes requested):
Significant build errors present. Please resolve these before requesting re-review.

I have added this driver to the build system via your FCF child PR. Please pull that up and test by invoking

```
cd app/rev3
make -jX # X = number of parallel processes you want for compilation
```

Build error log attached: [build_err_log.txt](https://github.com/user-attachments/files/29183796/build_err_log.txt)

MasterUser43:
Still have debug_assert to fix + maybe do more tinkering to avoid deprecated silly warnings. Will verify if effort is worth it or not (of warnings)

MasterUser43:
One simple note:

(1) When eventually config for rev3 is worked on:
```c
void HAL_SPI_TxRxCpltCallback(SPI_HandleTypeDef *hspi) {
    if (hspi == &(IMU_SPI)) {
        imu_process_async_cb();
    }
}

void HAL_SPI_ErrorCallback(SPI_HandleTypeDef *hspi) {
    if (hspi == &(IMU_SPI)) {
        imu_process_async_error_cb();
    }
}

```

ETSells:
Tagging @NArmistead on this as well, the more eyes we can get on big PRs like this the better

imu/imu_dflt.c
```diff
@@ -1,1282 +1,148 @@
...
-*       This software is licensed under terms that can be found in the LICENSE 
-*       file in the root directory of this software component.                 
-*       If no LICENSE file comes with this software, it is covered under the   
-*       BSD-3-Clause.                                                          
-*                                                                              
-*       https://opensource.org/license/bsd-3-clause          
-*
-*******************************************************************************/
-
+/**
```
ETSells, line 1: relates to: #46 -- imu portion complete under this CR, lets goooo

imu/imu.h
```diff
@@ -1,517 +1,51 @@
...
-typedef enum _IMU_FILTER_MODE
-    {
-    IMU_FILTER_FILTER_MODE  = ( 0x01 << 7 ),
-    IMU_FILTER_AVERAGE_MODE = ( 0x00 << 7 )
-    } IMU_FILTER_MODE;
+typedef struct {
+    float accel[3];         /* g                                   */
+    float gyro[3];          /* dps                                 */
+    float quaternion[4];    /* [w, x, y, z], unit quat              */
+    bool  quaternion_valid; /* false on legacy until AHRS warms up  */
```
ETSells, line 38: AHRS mentioned?????

imu/imu.h
```diff
@@ -1,517 +1,51 @@
...
+    bool  is_valid;
+} IMU_FlightData;
+
+typedef enum {
+    IMU_SYS_OK = 0,
+    IMU_SYS_FAIL,
+    IMU_SYS_BUSY,      /* async path: no new data yet */
+    IMU_SYS_NO_DATA,
+    IMU_SYS_INIT_FAIL
+} IMU_SYS_STATUS;
```
ETSells: Consider unifying the IMU_STATUS enum. As-is, the types are named the same between legacy and lsm and could break compatibility in shared modules between the two, unless IMU_SYS_STATUS is intended to replace it in the unified APIs. If that's the case, we'll want to do a follow-on to switch to the unified APIs soon.

imu/imu_lsm.h
```diff
@@ -0,0 +1,486 @@
...
+#define IMU_SPI_READ_BIT            ( 0x80U )
+
+/** @brief Blocking SPI timeout (ms) used during init and blocking reads       */
+#define IMU_SPI_TIMEOUT_MS          ( 10U )
+
+/**
+ * @brief SFLP game rotation vector ODR: 480 Hz (maximum supported).
+ *        DS14623 Rev3 Table 364: SFLP_GAME_ODR[2:0] = 101b
+ */
+#define IMU_SFLP_ODR                LSM6DSV320X_SFLP_480Hz
```
ETSells, line 85: 👀 has system integration implications

imu/imu_lsm.h
```diff
@@ -0,0 +1,486 @@
...
+ * @brief FIFO watermark threshold reference (programmed in FIFO slots/words).
+ *        1 LSB = 1 FIFO slot (7 bytes: 1-byte TAG + 6-byte data). DS14623 §9.8.
+ *
+ *        NOTE: This macro is a reference value; imu_init() configures the 
+ *        watermark dynamically based on the active dataset:
+ *          - Raw Base (Gyro + Accel):    2 slots
+ *          - SFLP Fusion Additions:     +3 slots (Quat, Grav, Gbias)
+ *
+ *        Rate Matching Behavior:
+ *          - Matched Rates (e.g., 480 Hz ODR / 480 Hz SFLP):
```
ETSells, line 97: we might wanna rate couple to this 480 hz tbh. depends on the performance impact of any data filtering we do, plus all of the I/O that the platform needs to support. Having a steady rate would fix a bunch of the architectural issues rev 2 deals with.

imu/imu_lsm.h
```diff
@@ -0,0 +1,486 @@
...
+ *            The FIFO accumulates exactly one synchronized dataset (5 slots) per interval.
+ *          
+ *          - Mixed Rates (e.g., 1.92 kHz ODR / 480 Hz SFLP):
+ *            The FIFO acts as an asynchronous reservoir. Raw data slots populate faster
```
ETSells, line 101: plus then we don't have to deal with emptying an async buffer and catching up our internal stestims to it.

@NArmistead, thoughts?

imu/imu_lsm.h
```diff
@@ -0,0 +1,486 @@
...
+
+/**
+ * @brief Data-ready poll timeout (ms) for imu_read_sync().
+ *        Must exceed the longest possible sample period. At the slowest
+ *        supported ODR (480 Hz), one period = 1000/480 = 2.083 ms, so 2 ms
+ *        would time out before the first sample arrives. 5 ms gives ~2.4x
+ *        headroom at 480 Hz and ~9.6x at the default 1920 Hz.
+ *        DS14623 Rev3 Table 54.
+ */
+#define IMU_DRDY_TIMEOUT_MS         ( 5U )
```
ETSells: note (but please don't feel obligated to fix): timers are specified in systicks (1ms) unless otherwise noted. _MS is assumed. it's the same reason we don't specify things like imu_acc_x_mps -- SI units are expected internally unless otherwise noted.

imu/imu_lsm.h
```diff
@@ -0,0 +1,486 @@
...
+ */
+typedef struct _IMU_RAW
+    {
+    int16_t accel_x;   /* Accelerometer X raw counts */
+    int16_t accel_y;   /* Accelerometer Y raw counts */
+    int16_t accel_z;   /* Accelerometer Z raw counts */
+    int16_t gyro_x;    /* Gyroscope X raw counts     */
+    int16_t gyro_y;    /* Gyroscope Y raw counts     */
+    int16_t gyro_z;    /* Gyroscope Z raw counts     */
+    } IMU_RAW;
```
ETSells, line 178: same as legacy lol, nice

imu/imu_lsm.h
```diff
@@ -0,0 +1,486 @@
...
+
+/**
+ * @brief SFLP sensor fusion output (float32).
+ *        Quaternion is the game rotation vector (6-axis, no magnetometer).
```
ETSells, line 182: maybe we should have put the magnetometer on this device's aux, if it has it. we've always known that capability exists but we've never cared. it'd be super helpful now though.

imu/imu_lsm.h
```diff
@@ -0,0 +1,486 @@
...
+    float gbias_z;     /* Gyroscope bias Z                                   */
+    } IMU_SFLP_DATA;
+
+/**
+ * @brief Physical-unit converted IMU output.
+ *        Sensitivity values from DS14623 Rev3 Table 3.
+ */
+typedef struct _IMU_DATA
+    {
+    float accel_x;     /* Accelerometer X [g]   */
```
ETSells: cr: please use m/s^2 internally, we expect SI units. should be as simple as applying a 9.8 m/s^2 scalar, or removing the inverse if you're converting to g yourself.

imu/imu_lsm.h
```diff
@@ -0,0 +1,486 @@
...
+ * @brief Gyroscope full-scale selection.
+ *        DS14623 Rev3 Table 65: FS_G[2:0] in CTRL6 (15h).
+ */
+typedef enum _IMU_GYRO_FS
+    {
+    IMU_GYRO_FS_250DPS  = LSM6DSV320X_250dps,   /* ±250  dps,  8.75  mdps/LSB */
+    IMU_GYRO_FS_500DPS  = LSM6DSV320X_500dps,   /* ±500  dps, 17.50  mdps/LSB */
+    IMU_GYRO_FS_1000DPS = LSM6DSV320X_1000dps,  /* ±1000 dps, 35.0   mdps/LSB */
+    IMU_GYRO_FS_2000DPS = LSM6DSV320X_2000dps,  /* ±2000 dps, 70.0   mdps/LSB */
+    IMU_GYRO_FS_4000DPS = LSM6DSV320X_4000dps,  /* ±4000 dps, 140.0  mdps/LSB */
```
ETSells, line 270: leaning towards the following for rev 3 app integration:

- idle in 500 dps and 16G, low-power mode (4 avg acc, lp for gyro?)
- calibrate in ~500 dps and 16G for resolution, normal mode
- switch to 4000 dps and 128G for launch detect, high performance
- drop back to 16G and 1000 DPS for coast and beyond, high performance or normal mode

either of you have concerns about these choices?

imu/imu_dflt.c
ETSells, line 1: cr: This forces hyperspecific choices for configuration when they should be project dependent. Other than that though, I really like this idea. Consider renaming this file to "imu_dflt.c" and allowing a project to define its own implementations of these functions as contract functions, that way projects aren't forced into this implementation. 

You can also specify example DMA completion callbacks in here so projects (me and @NArmistead) know how to integrate this new form.

ETSells: imu.h can stay as is since the enum and function prototypes should stay common, it's only the implementation that will change.

imu/imu_lsm.c
```diff
@@ -0,0 +1,1516 @@
...
+
+/** @brief Number of FIFO slots captured in the most recent DMA transaction    */
+static volatile uint8_t imu_dma_slots_requested;
+
+/**
+ * @brief Synchronization flags.
+ *        volatile: both are written in interrupt context (imu_process_async_cb)
+ *        and read from task/loop context.
+ */
+static volatile bool imu_dma_busy  = false;
```
ETSells: nit: use atomic_bool from <stdatomic.h>. This will work fine 99.99999% of the time, but it is technically possible to receive and handle an interrupt signal in the middle of a read-modify-write operation. That'd be the most annoying race condition possible lol

imu/imu_lsm.c
```diff
@@ -0,0 +1,1516 @@
...
+    IMU_ODR odr
+    )
+{
+switch ( odr ) {
+    case IMU_ODR_480HZ:  return LSM6DSV320X_ODR_AT_480Hz;
+    case IMU_ODR_960HZ:  return LSM6DSV320X_ODR_AT_960Hz;
+    case IMU_ODR_3840HZ: return LSM6DSV320X_ODR_AT_3840Hz;
+    case IMU_ODR_7680HZ: return LSM6DSV320X_ODR_AT_7680Hz;
+    case IMU_ODR_1920HZ:
+    default:             return LSM6DSV320X_ODR_AT_1920Hz;
```
ETSells: nit: consider debug assert here like: `imu_assert( odr == IMU_ODR_1920HZ )` to catch misconfig. not strictly necessary but good practice

imu/imu_lsm.c
```diff
@@ -0,0 +1,1516 @@
...
+        case IMU_ACC_FS_8G:  xl_fs = LSM6DSV320X_8g;  break;
+        case IMU_ACC_FS_16G: xl_fs = LSM6DSV320X_16g; break;
+        case IMU_ACC_FS_2G:  /* fall-through */
+        default:             xl_fs = LSM6DSV320X_2g;  break;
+    }
+    pid_status |= lsm6dsv320x_xl_full_scale_set( &imu_ctx, xl_fs );
+}
+
+/* Gyro FS: map IMU_GYRO_FS to lsm6dsv320x_gy_full_scale_t. DS14623 Table 65. */
+{
```
ETSells: cr: what's the condition here? getting hard to track

imu/imu_lsm.c
ETSells, line 1: whole file: SDR bracing style not followed.

imu/imu_dflt.c
ETSells, line 1: whole file: SDR bracing style not followed.

imu/imu_lsm.c
```diff
@@ -0,0 +1,1516 @@
...
+
+
+/**
+  * @brief  Converts three half-float SFLP components to a unit quaternion.
+  * @note   ST packs only X, Y, Z into a TAG=0x13 FIFO slot; W is defined:
+  *         w = sqrt(1 − x² − y² − z²) [unit quaterinion]
+  *         If numerical noise causes |xyz|² > 1, XYZ is normalised first so
+  *         the output remains a valid unit quaternion.
+  *         Uses lsm6dsv320x_from_quaternion_lsb_to_float() 
+  * @param  quat:  Output [x, y, z, w] indexed [0..3].
```
ETSells: cr: our internal format is [w, x, y, z]. would prefer that to stay consistent so we don't have to remap every time this is called, unless you already do that in the caller.

imu/imu_lsm.c
```diff
@@ -0,0 +1,1516 @@
...
+    case IMU_FIFO_TAG_SFLP_GBIAS:
+        if ( sflp_ptr != NULL ) {
+            sflp_ptr->gbias_x = lsm6dsv320x_from_quaternion_lsb_to_float( x_raw );
+            sflp_ptr->gbias_y = lsm6dsv320x_from_quaternion_lsb_to_float( y_raw );
+            sflp_ptr->gbias_z = lsm6dsv320x_from_quaternion_lsb_to_float( z_raw );
+        }
+        break;
+
+    default:
+        /* Unknown or unhandled tag - silently skip */
```
ETSells: i wouldn't silently skip, consider defensive programming via debug assert

imu/imu_lsm.c
```diff
@@ -0,0 +1,1516 @@
...
+slots_to_read = ( (uint16_t)fifo_status.fifo_level < (uint16_t)IMU_DMA_BUF_SLOTS )
+              ? (uint8_t)fifo_status.fifo_level
+              : (uint8_t)IMU_DMA_BUF_SLOTS;
+
+/*
+ * Invariant: clamped slot count must fit within the fixed DMA buffer.
+ * Fires if IMU_DMA_BUF_SLOTS was reduced below fifo_status.fifo_level
+ * without a corresponding buffer resize - a build-time mistake.
+ */
+IMU_ASSERT( slots_to_read <= IMU_DMA_BUF_SLOTS );
```
ETSells: if it's a build-time mistake, you could _Static_assert?

imu/imu_lsm.c
```diff
@@ -0,0 +1,1516 @@
...
+/*
+ * Critical section: snapshot the slot count, derive exactly how many bytes
+ * this DMA transaction actually populated (1 cmd byte + slots*7), copy only
+ * that range, then clear the ready flag. The slot count MUST be captured
+ * before it's used to size the copy (to avoid garbage/wrapping)
+ * memcpy protection req. + the parse runs outside the critical
+ * section so FPU operations do not block other interrupts.
+ */
+primask = __get_PRIMASK();
+__disable_irq();
```
ETSells: rather than a global interrupt disablement like this, we tend to prefer targeted calls. this is fine, but just keep in mind that this also disables timer interrupts, interrupts for other sensors, externally driven interrupts, etc.

ETSells: cr: critical section never exited with irq re-enablement. would break the system entirely

ETSells (changes requested):
Couple CRs, lots of nits. This is absolutely sweet though, and i love the default APIs you wrote in imu.c -- should make it very easy to use for integrating projects.

imu/imu.h
NArmistead: what!!

imu/imu_lsm.c
```diff
@@ -0,0 +1,1516 @@
...
+static IMU_ODR imu_cached_odr;
+static bool    imu_cached_fifo_en;
+
+/**
+ * @brief Cached power mode most recently requested via imu_init() /
+ *        imu_set_power_mode(). DS14623 Rev3 §6.1.2 forces the low-G chain
+ *        into high-performance mode whenever the high-G chain is active;
+ *        imu_set_accel_fs() uses this cache to restore the user's actual
+ *        requested mode after a High-G -> Low-G transition removes that
+ *        constraint, instead of leaving the sensor stuck in HP mode.
```
NArmistead, line 127: nit: arguably not very brief. consider moving everything but the first sentence to a new paragraph

imu/imu_lsm.c
MasterUser43: Weirdo

imu/imu.h
```diff
@@ -1,517 +1,51 @@
...
-/* Filter Mode */
-typedef enum _IMU_FILTER_MODE
-    {
-    IMU_FILTER_FILTER_MODE  = ( 0x01 << 7 ),
-    IMU_FILTER_AVERAGE_MODE = ( 0x00 << 7 )
-    } IMU_FILTER_MODE;
+typedef struct {
+    float accel[3];         /* g                                   */
+    float gyro[3];          /* dps                                 */
+    float quaternion[4];    /* [w, x, y, z], unit quat              */
```
NArmistead, line 37: This can be replaced by the QUAT struct once the new state estimation changes are merged.  It'll be the exact same in memory.

The other issue is that on legacy this quaternion is part of the state estimation struct and is calculated from IMU values in the sensor system. It's meant to be less coupled to the IMU driver directly because the IMU is only giving accel and gyro.

imu/imu_lsm.h
```diff
@@ -0,0 +1,486 @@
...
+ * @brief SFLP game rotation vector ODR: 480 Hz (maximum supported).
+ *        DS14623 Rev3 Table 364: SFLP_GAME_ODR[2:0] = 101b
+ */
+#define IMU_SFLP_ODR                LSM6DSV320X_SFLP_480Hz
+
+/**
+ * @brief FIFO watermark threshold reference (programmed in FIFO slots/words).
+ *        1 LSB = 1 FIFO slot (7 bytes: 1-byte TAG + 6-byte data). DS14623 §9.8.
+ *
+ *        NOTE: This macro is a reference value; imu_init() configures the
```
NArmistead: nit: you can use `@note` here

imu/imu_lsm.h
NArmistead: Do we get noticeably better accuracy at a super high ODR? 480 Hz is already pretty fast and its probably higher than our frame rate (I don't really know on rev3) so we won't have any issues like what we have with baro on rev2. I think this setting would be continuous-to-FIFO mode but I don't see where we would want to change the FIFO mode

imu/imu_lsm.c
MasterUser43: 1405: __set_PRIMASK( primask ); 

But the _disable_irq() is a bit bothersome, I'll see if it's worth NVIC disabling. 

I could also do double-buffering but...

Just notes for myself.

imu/imu_lsm.h
MasterUser43: Yeah it's possible, it's more of a hardware routing thing tbh.

ETSells: i dont think we did that routing is the thing. we have it on one of the MCU's I2C buses

imu/imu_legacy.c
```diff
@@ -0,0 +1,1284 @@
...
+if ( imu_status != IMU_OK )
+	{
+	return imu_status;
+	}
+
+/* Combine high byte and low byte to 16 bit data  */
+#if   defined( A0002_REV1 )
+    gyro_x_raw = ( (uint16_t) regGyro[0] ) << 8 | regGyro[1];
+    gyro_y_raw = ( (uint16_t) regGyro[0] ) << 8 | regGyro[1];
+    gyro_z_raw = ( (uint16_t) regGyro[0] ) << 8 | regGyro[1];
```
MasterUser43: I assume it is meant to be reading [2], [3] & [4], [5] respectively.

Not touched b/c legacy code, haven't combed through fully.
```suggestion
    gyro_y_raw = ( (uint16_t) regGyro[2] ) << 8 | regGyro[3];
    gyro_z_raw = ( (uint16_t) regGyro[4] ) << 8 | regGyro[5];
```

MasterUser43: Ooh fancy codium extension :). Just wanted to test it.

imu/imu_lsm.h
MasterUser43: Are we using the BMM350?

Looking more into the datasheet of the IMU, connecting to the aux seems like it just provides FIFO synchronization at the cost of configuring the LSM I2C master, so direct bus debugging is gonna be fun there.

Additionally, because the LSM's on-chip SFLP engine is 'strictly' a 6-axis fusion block anyway, 9-axis absolute orientation has to be computed on the MCU. 

Prob better to keep it as a separate mag_xx.c/h files plus more independent control anyway.

Rev 4 board problem hehe.

imu/imu_legacy.c
ETSells: oh oof why was this wrong in legacy?

cc: @narmistead

NArmistead:
I can’t figure out how to reply without it being part of the review but the gyro reading is rev1 which is why I didn’t touch it but yeah I think it’s wrong

imu/imu_legacy.c
ETSells: Image

OOF bro this has been borked for a looooong time

ETSells: Feel free to fix, but this is not compiled into rev 2 as it is for an unsupported platform.

imu/imu_legacy.c
```diff
@@ -415,8 +418,8 @@ if ( imu_status != IMU_OK )
 /* Combine high byte and low byte to 16 bit data  */
 #if   defined( A0002_REV1 )
     gyro_x_raw = ( (uint16_t) regGyro[0] ) << 8 | regGyro[1];
-    gyro_y_raw = ( (uint16_t) regGyro[0] ) << 8 | regGyro[1];
-    gyro_z_raw = ( (uint16_t) regGyro[0] ) << 8 | regGyro[1];
+    gyro_y_raw = ( (uint16_t) regGyro[2] ) << 8 | regGyro[3];
+    gyro_z_raw = ( (uint16_t) regGyro[4] ) << 8 | regGyro[5];
```
NArmistead: You might be able to just remove this because it's rev1 code (@ETSells ?)

imu/imu_lsm.h
NArmistead: Vertical acceleration during coast may be above 16G especially for higher power rockets. It's only for a couple seconds but that's long enough to be a problem

imu/imu_lsm.c
```diff
@@ -0,0 +1,1549 @@
...
+
+
+/**
+  * @brief  Converts three half-float SFLP components to a unit quaternion.
+  * @note   ST packs only X, Y, Z into a TAG=0x13 FIFO slot; W is defined:
+  *         w = sqrt(1 − x² − y² − z²) [unit quaterinion]
+  *         If numerical noise causes |xyz|² > 1, XYZ is normalised first so
+  *         the output remains a valid unit quaternion.
+  *         Uses lsm6dsv320x_from_quaternion_lsb_to_float() 
+  * @param  quat:  Output [w, x, y, z] indexed [0..3].
```
NArmistead: super nit: usually there isnt a `:` after the name of the param

NArmistead (commented):
Sorry I don't have time to fully review everything (mostly didn't get to imu_lsm.c) but I want to get this out before I'm gone for the rest of the week.

ETSells:
> Do we get noticeably better accuracy at a super high ODR? 480 Hz is already pretty fast and its probably higher than our frame rate (I don't really know on rev3) so we won't have any issues like what we have with baro on rev2. I think this setting would be continuous-to-FIFO mode but I don't see where we would want to change the FIFO mode

I think my plan is to couple system frame rate to the odr of the slowest main sensor (baro, accel, gyro, and state estim fusion). I dont think I would care about changing the FIFO because we'll pretty much always just want to take the most recent reading from the device.

imu/imu_legacy.c
ETSells: Yes, safe to remove.

ETSells:
> Vertical acceleration during coast may be above 16G especially for higher power rockets. It's only for a couple seconds but that's long enough to be a problem

during coast? Under power it can go up to ~120 ish Gs in high power mode, but coast should be pure drag + gravity, which shouldn't be that high. We can add a decel phase to the fsm if we expect a really high instantaneous slowdown?

imu/imu_lsm.c
ETSells: that meme Robert made about me never approving pull requests will be true for you too, just you wait

NArmistead:
> > Do we get noticeably better accuracy at a super high ODR? 480 Hz is already pretty fast and its probably higher than our frame rate (I don't really know on rev3) so we won't have any issues like what we have with baro on rev2. I think this setting would be continuous-to-FIFO mode but I don't see where we would want to change the FIFO mode
> 
> I think my plan is to couple system frame rate to the odr of the slowest main sensor (baro, accel, gyro, and state estim fusion). I dont think I would care about changing the FIFO because we'll pretty much always just want to take the most recent reading from the device.

Okay, but if we do a kalman filter we can update the state estimation as data from each sensor arrives instead of coupling it to the slowest (e.g. 10 updates with baro then one with GPS).

ETSells:
> > > Do we get noticeably better accuracy at a super high ODR? 480 Hz is already pretty fast and its probably higher than our frame rate (I don't really know on rev3) so we won't have any issues like what we have with baro on rev2. I think this setting would be continuous-to-FIFO mode but I don't see where we would want to change the FIFO mode
> > 
> > I think my plan is to couple system frame rate to the odr of the slowest main sensor (baro, accel, gyro, and state estim fusion). I dont think I would care about changing the FIFO because we'll pretty much always just want to take the most recent reading from the device.
> 
> Okay, but if we do a kalman filter we can update the state estimation as data from each sensor arrives instead of coupling it to the slowest (e.g. 10 updates with baro then one with GPS).

lol nvm thats definitely easier from an architectural perspective, id prefer that

NArmistead:
> > Vertical acceleration during coast may be above 16G especially for higher power rockets. It's only for a couple seconds but that's long enough to be a problem
> 
> during coast? Under power it can go up to ~120 ish Gs in high power mode, but coast should be pure drag + gravity, which shouldn't be that high. We can add a decel phase to the fsm if we expect a really high instantaneous slowdown?

Drag deceleration for a sim I have with a K400 motor gets above 50G

ETSells:
damn, okay. so we probably want a mechanism to track deceleration to subsonic in between boost and coast? bc that decel probably also requires a decently high frame rate to capture the acceleration, whereas coast can go slower and descent under parachutes can go even slower (if all that is needed)

imu/imu_dflt.c
ETSells: Still true.

imu/imu_lsm.c
```diff
@@ -131,24 +133,41 @@ static IMU_ACC_MODE imu_cached_acc_mode;
...
+ */
+static uint8_t imu_dma_fill_idx = 0U;
 
-/** @brief Number of FIFO slots captured in the most recent DMA transaction    */
-static volatile uint8_t imu_dma_slots_requested;
+/**
+ * @brief Index of the imu_dma_rx_buf row holding the most recently completed,
+ *        not-yet consumed transfer. 
+ */
+static atomic_uchar imu_dma_ready_idx = 0U;
```
ETSells: nit: we prefer not using types like 'char' and 'int' in favor of types that are explicit about the size they contain like 'uint8_t'. given that atomics only have so many aliases this might be fine here tho?

imu/imu_lsm.c
ETSells: still true

ETSells, line 387: i'm with nick -- we should consider using the math_sdr.h QUAT type once it exists.

imu/imu_lsm.c
```diff
@@ -491,9 +521,9 @@ switch ( tag ) {
   *        it from the flight loop.
   *
   * @param  config: Pointer to user configuration struct (non-NULL).
-  * @retval IMU_STATUS
+  * @retval IMU_LSM_STATUS
   */
-IMU_STATUS imu_init
+IMU_LSM_STATUS imu_init
```
ETSells, line 569: note: this is gonna force us to use IMU_SYS_STATUS for anything IMU related that isn't platform dependent. we should prepare to make this switch during initial rev 3 integration.

imu/imu_lsm.c
```diff
@@ -131,24 +133,41 @@ static IMU_ACC_MODE imu_cached_acc_mode;
  *        - Section (.dma_buffer): Forces placement in RAM_D2 (0x30000000).
  *          DTCM (.bss) is not reachable by the DMA1/DMA2 controllers.
```
ETSells, line 135: man, that's unfortunate. DTCM doesn't cache, so its what we would *want* to use otherwise

imu/imu_lsm.c
```diff
@@ -1234,30 +1268,22 @@ imu_dma_tx_buf[0] = (uint8_t)( LSM6DSV320X_FIFO_DATA_OUT_TAG | IMU_SPI_READ_BIT
...
- *     fetches speculatively during the transfer.
+/* 
+ * Maintain Cortex-M7 D-Cache coherency for DMA buffers:
+ *   TX: Clean (flush) the TX buffer so the DMA peripheral reads the written SPI command.
+ *   RX: Invalidate the active RX buffer to discard any stale/speculatively prefetched data.
  */
 SCB_CleanDCache_by_Addr( (uint32_t*)imu_dma_tx_buf,
                           (int32_t)IMU_DMA_BUF_BYTES_ALIGNED );
-SCB_InvalidateDCache_by_Addr( (uint32_t*)imu_dma_rx_buf,
+SCB_InvalidateDCache_by_Addr( (uint32_t*)imu_dma_rx_buf[ imu_dma_fill_idx ],
```
ETSells, line 1438: caching on our mcu is so weird... i originally was concerned at the idea of needing to invalidate each region separately since I thought they'd be in the same page, but there are no pages on the M7. how does caching even work lmao

imu/imu_lsm.c
```diff
@@ -1381,29 +1411,22 @@ IMU_ASSERT( raw != NULL || sflp != NULL );
...
- * that range, then clear the ready flag. The slot count MUST be captured
- * before it's used to size the copy (to avoid garbage/wrapping)
- * memcpy protection req. + the parse runs outside the critical
- * section so FPU operations do not block other interrupts.
+ * Retrieve the active row index. The double-buffering scheme and the DMA 
+ * busy-gate guarantee the DMA controller will not overwrite this row 
+ * while copying it.
  */
-primask = __get_PRIMASK();
-__disable_irq();
```
ETSells: ah the other concern i forgot to mention with this is that it'll already be sitting in an IRQ disabled critical section when it's called, so double disablement of IRQs would fuck us over.

imu/imu_lsm.c
```diff
@@ -1443,6 +1466,11 @@ float   gyro_sens;
 uint8_t acc_idx;
 uint8_t gy_idx;
 
+/** @brief Standard gravity, used to convert g -> m/s^2 (SI). */
+#define IMU_STANDARD_GRAVITY        ( 9.80665f )
```
ETSells: these should already be defined as constants somewhere, right @NArmistead?

imu/imu_lsm.c
```diff
@@ -1500,15 +1528,20 @@ switch ( imu_gyro_fs ) {
 IMU_ASSERT( acc_idx < 9U );
 IMU_ASSERT( gy_idx  < 5U );
 
-acc_sens  = acc_sens_mg_lsb[acc_idx]   / 1000.0f; /* mg/LSB  -> g/LSB   */
-gyro_sens = gyro_sens_mdps_lsb[gy_idx] / 1000.0f; /* mdps/LSB -> dps/LSB */
+/* mg/LSB -> g/LSB -> (SI) m/s^2/LSB   */
+acc_sens  = ( acc_sens_mg_lsb[acc_idx]   / 1000.0f ) * IMU_STANDARD_GRAVITY;
+/* mdps/LSB -> dps/LSB -> (SI) rad/s/LSB */
+gyro_sens = ( gyro_sens_mdps_lsb[gy_idx] / 1000.0f ) * IMU_DEG_TO_RAD;
```
ETSells: cr: expected system units for rotation are degrees per second.

ETSells (changes requested):
one CR remaining.

remind me to enumerate sdr coding standards somewhere in like a markdown file, i have no way of telling people what they are except showing them.

imu/imu_lsm.c
NArmistead: should be in math_sdr.h

NArmistead: although the precision on that is super low. not sure why i did that

ETSells: http://jpl.nasa.gov/edu/news/how-many-decimals-of-pi-do-we-really-need/

its probably fine but we could def stand to get a bit more precise lol

imu/imu_lsm.c
MasterUser43: Not a yap enjoy-er I see

imu/imu_lsm.c
MasterUser43: I like the attention, dw

imu/imu_lsm.c
MasterUser43: Ye, b/c of that M7 lack of cache snooping for DMA, basically had to manually sync the L1 Cache in code since DTCM is private :(. D2-SRAM (`.dma_buffer`) is behind the L1-Cache, so yoink that bad boi there.

```c
__ALIGNED(32) __attribute__((section(".dma_buffer")))
static uint8_t imu_dma_rx_buf[ 2 ][ IMU_DMA_BUF_BYTES_ALIGNED ];
```
Data is managed in 32-byte cache lines instead of the 4kB pages. Better for precision and efficiency, but no hardware snooping by design. Usually I just round to ensure no neighborhood destruction [false sharing according to google] & ptr reference correctly.

This section you highlighted basically flags CPU's internal cache. The clean (TX) forces 'write-back' cache to flush SPI cmd into RAM (DMA-readable). The invalidate (RX), flags current RX buffer as stale requiring re-fetching from RAM post DMA-write.

My understanding for this is just treat DMA as external master relative to the L1-Cache. DMA don't know CPU has modified data (in Cache) nor does the CPU know DMA has modified data (in RAM). The code just gotta be the mailman to let them know when to update.

imu/imu.h
MasterUser43: https://github.com/SunDevilRocketry/mod/pull/134

This guy I assume?

ETSells: Yep, exactly. Nick and I need to cross-review each other's changes, and I also need our software lead to approve my dashboard PR before these all get merged.

imu/imu_lsm.c
MasterUser43: Honestly was keeping distinction because I never was going to touch `imu_legacy.c`, but tbh for measurement SI consistency, turns out I'm going to end up touching those files anyway, so probably will go back to `IMU_STATUS` and just unify 'em.

imu/imu_lsm.c
MasterUser43: Very SI of you.

imu/imu_legacy.c
MasterUser43: Should I nuke all rev1 code from legacy then?

imu/imu_lsm.c
ETSells: tbh we *should* do radians, but oh well. I would rather not change it since it feeds into state estimation algorithms that would have to change

imu/imu_legacy.c
ETSells: yeah feel free to yeet. we'll be testing the rev 2 path for this prior to integration/merging, so go ahead. if it breaks smth ill lyk

imu/imu_lsm.c
NArmistead: Ngl converting to rads would be fine because all of the state estimation stuff needs to convert from degrees to rads anyway except for the final roll rate

ETSells: Okay, in that case feel free to leave it and we'll backport rev 2. Should be no loss of precision bc floating points are sick

MasterUser43: Ofc, tell me after I put things in dps lmao

ETSells: PFFT- okay well dont go through the trouble to revert now that you've done it

MasterUser43:
Tempted to just add
```c
\
-DA0002_REV2
```
to the makefile b/c sensor.h making the gps scream in agony lmao.

```c
#if !defined( A0002_REV2 )
#error "imu_legacy.h is only supported on the A0002_REV2 hardware platform"
#endif
```

Ya'll think I should just remove the header or just fix the makefile.

imu/imu_lsm.c
MasterUser43: Actually, don't know if I would want to unify them because of `IMU_MAG_INIT_FAIL` wouldn't make sense to be with the new rev3 IMU. Plus, legacy is I2C & new is SPI, so coupling a unified with both I2C and SPI errors is polluting 

If you think the name is the annoying part, I can just rename `IMU_STATUS` on legacy side into `IMU_LEGACY_STATUS` and rename the unifer `IMU_SYS_STATUS` to `IMU_STATUS`. 

So honestly, I'm going to wait for the `QUAT` PR and revist this later. Point of the ```imu_dflt.c``` is to be the adapter anyway.

ETSells: Yep that works! I wasn't critiquing imu_sys_status, just noting for myself that it changes app integration considerations since imu_status becomes platform specific. as-is is fine

ETSells:
Definitely don't define something that isnt true. If youre fixing it, fix the root cause. To do otherwise is to accept technical debt, and that just makes our lives harder down the road. Especially early in a system's lifetime, we should be really smart about the tech debt we introduce.

MasterUser43:
I mean, it feels odd just yoinking that random include. Seems like it doesn't do anything for GPS testing? Lmk

MasterUser43:
Actually might be a poptart (bad) fix... yea, Ill see to this another day.

Am actually curious why GPS is asking for sensor.h anyway. Just gonna fix my guard nonetheless b/c yucky.

_test/driver/gps/test_gps.c
```diff
@@ -23,7 +23,6 @@ Project Includes
 ------------------------------------------------------------------------------*/
 #include "sdrtf_pub.h"
 #include "main.h"
```
MasterUser43, line 25: This file.

ETSells:
I think GPS shouldn't have a dependency on sensor.h. That guard might have originally existed from me trying to get the STM32F7 based ground station to compile with the telemetry.h header. I did a lot of sketchy shit for it.

ETSells:
> Actually might be a poptart (bad) fix... yea, Ill see to this another day.
> 
> Am actually curious why GPS is asking for sensor.h anyway. Just gonna fix my guard nonetheless b/c yucky.

what do you have against pop tarts :(

MasterUser43:
Gonna leave the GPS unmodified b/c out of scope anyway. Make a note for future updates to the test b/c random footnote imu touching test is quite silly organizationally.

Wonder how long this PR is going get lmao.

imu/imu_legacy.c
```diff
@@ -297,19 +300,13 @@ HAL_NVIC_EnableIRQ(I2C2_EV_IRQn);
...
-*                                                                              *
-* DESCRIPTION:                                                                 * 
-* 		Return the pointer to structure that updates the                       * 
-*       x,y,z acceleration values from the IMU                                 *
-*                                                                              *
-*******************************************************************************/
+/**
+  * @brief  Blocking read of accelerometer X/Y/Z.
+  * @param  pIMU: Destination struct for the accel readout (size: 12 bytes).
+  * @retval IMU_STATUS
```
ETSells: nit: doxygen "retval" vs "return" semantics

imu/imu_legacy.c
```diff
@@ -612,56 +547,54 @@ return imu_status;
...
+  * @brief  Initialize the magnetometer.
+  *
+  * @note   This function is heavily derived from the official Bosch BMM150
+  *         driver, which is protected by the BSD-3-Clause license. This
+  *         function is exempt from any licensing that may be applied to a
+  *         current/future Sun Devil Rocketry project. Per the terms of the
+  *         BSD-3-Clause license, the following notice is retained from the
+  *         source project and applies to the procedure below.
+  *
+  *         Copyright (c) 2020 Bosch Sensortec GmbH. All rights reserved.
```
ETSells: nit: "@copyright" tag instead of "note". this is fine in accordance with our licensing policies but for documentation generation itd be nice to have this show up in the right section.

imu/imu_legacy.c
```diff
@@ -1184,15 +1081,12 @@ else if( imu_data_ready && !mag_data_ready )
...
-*       Getter function for IMU sensor data (IT).                              *
-*                                                                              *
-*******************************************************************************/
+/**
+  * @brief  Getter function for IMU sensor data (interrupt mode).
+  * @param  cpy_ptr: Destination struct for the copied sensor readout.
+  * @retval IMU_STATUS: IMU_BUSY until both imu_data_ready and mag_data_ready
+  *         are set by imu_it_handler().
+  */
 IMU_STATUS get_imu_it
```
ETSells, line 1114: observation (i will do this myself at some point): I've begun to push for an unofficial policy that all non-static functions defined in a module should be preceded by that module's name. For example, instead of print_debug we'd say debug_print.

This convention also applies to private functions within a module of multiple files. See the test framework for an example, where we preface private functions that need to be accessed by the other framework file with _test.

The idea for this is to help with intellisense autocompletion -- if you need to call a function from a module but you're not sure what it is, you should be able to type {module}_{keyword} and have your autocompletion tool come up with the right function, e.g. imu_accel should present imu_get_accel_and_gyro as an autocomplete option. Haven't tested this a ton with vscode yet, but that kind of convention works great for jetbrains IDEs & visual studio.

imu/imu_lsm.c
```diff
@@ -0,0 +1,1546 @@
...
+switch ( odr ) {
+    case IMU_ODR_480HZ:  return LSM6DSV320X_ODR_AT_480Hz;
+    case IMU_ODR_960HZ:  return LSM6DSV320X_ODR_AT_960Hz;
+    case IMU_ODR_3840HZ: return LSM6DSV320X_ODR_AT_3840Hz;
+    case IMU_ODR_7680HZ: return LSM6DSV320X_ODR_AT_7680Hz;
+    case IMU_ODR_1920HZ: return LSM6DSV320X_ODR_AT_1920Hz;
+    default:
+        IMU_ASSERT( odr == IMU_ODR_1920HZ );
+        return LSM6DSV320X_ODR_AT_1920Hz;
+}
```
ETSells: nit: sdr bracing style

imu/imu_lsm.c
ETSells: still true. i know our coding style is a little weird, but this uniformity makes it a lot easier to read

imu/imu_lsm.c
```diff
@@ -0,0 +1,1546 @@
...
+imu_ctx.handle    = NULL;
+
+
+/*--------------------------------------------------------------------------
+ Implementation
+--------------------------------------------------------------------------*/
+
+/* Disable INT1 EXTI to prevent accidental ISR triggers during init.
+   INT1 is routed to EXTI4 (PC4 = IMU_INT1_PIN). */
+HAL_NVIC_DisableIRQ( EXTI4_IRQn );
```
ETSells: This is platform specific, so we may want to define it in the pindefs file

ETSells: Also, we might want to disable interrupts on the SPI handle as well, although it should be fine as long as we use tx/rx cplt callbacks.

imu/imu_lsm.c
```diff
@@ -0,0 +1,1546 @@
...
+__NOP(); __NOP();
+HAL_GPIO_WritePin( IMU_NSS_GPIO_PORT, IMU_NSS_PIN, GPIO_PIN_SET );
+
+if ( !imu_dma_busy ) { return; }
+
+/* Invalidate D-Cache lines covering the row that was just DMA'd into */
+SCB_InvalidateDCache_by_Addr( (uint32_t*)imu_dma_rx_buf[ imu_dma_fill_idx ],
+                               (int32_t)IMU_DMA_BUF_BYTES_ALIGNED );
+
+/* Publish the ready index before setting the ready flag to prevent race conditions */
```
ETSells, line 1499: shouldnt happen unless a higher priority ISR tries to consume this data, which we would not allow architecturally. still a nice protective touch.

ETSells (approved):
yuh

still some issues with adherence to standards but i can fix myself if/when i resolve my observation

imu/imu_legacy.c
MasterUser43: Would have to scrap through all the stuff that relies on imu_legacy lmao

imu/imu_lsm.c
MasterUser43: It depends on whether we need to support mid-flight re-initialization for error recovery, but I’ve added `HAL_SPI_Abort()` and a manual CS release to the top of the function anyway. Better than disabling IRQ so SPI isn't locked out by a low CS pin.

MasterUser43: Should be updated in lib.

imu/imu_lsm.c
MasterUser43: Yeee

ETSells (approved):
I only took a small sample of this PR, but everything looks like it follows style guidelines now too! fuck it we ball, ship it

imu/imu_legacy.c
ETSells: yeah this is not me telling you to do it, it's me saying I'm gonna do it at some point

MasterUser43: Well, ye, a glhf to you :P

MasterUser43:
image

PR #31, opened 2026-05-20
Change raw IMU to signed int

NArmistead:
## Description
`uint16_t` → `int16_t` for accel and gyro values in `IMU_RAW`

### Testing
- [ ] Unit tests modified

### Other

### Standards
- [x] Code is testable

### Error Handling

### Memory

### Performance
- [x] Busy waiting is avoided

### DIFF
--- a/imu/imu.c
+++ b/imu/imu.c
@@ -384,9 +384,9 @@ IMU_STATUS imu_get_gyro_xyz
  Local variables 
 ------------------------------------------------------------------------------*/
 uint8_t     regGyro[6];   /* Bytes from gyro registers */
-uint16_t    gyro_x_raw;   /* Raw gyro sensor readouts  */
-uint16_t    gyro_y_raw; 
-uint16_t    gyro_z_raw; 
+int16_t    gyro_x_raw;    /* Raw gyro sensor readouts  */
+int16_t    gyro_y_raw; 
+int16_t    gyro_z_raw; 
 IMU_STATUS  imu_status;   /* IMU status return codes   */
 
 
@@ -417,9 +417,9 @@ if ( imu_status != IMU_OK )
     gyro_y_raw = ( (uint16_t) regGyro[0] ) << 8 | regGyro[1];
     gyro_z_raw = ( (uint16_t) regGyro[0] ) << 8 | regGyro[1];
 #elif defined( A0002_REV2 )
-    gyro_x_raw = ( (uint16_t) regGyro[1] ) << 8 | regGyro[0];
-    gyro_y_raw = ( (uint16_t) regGyro[3] ) << 8 | regGyro[2];
-    gyro_z_raw = ( (uint16_t) regGyro[5] ) << 8 | regGyro[4];
+    gyro_x_raw = ( (int16_t) regGyro[1] ) << 8 | regGyro[0];
+    gyro_y_raw = ( (int16_t) regGyro[3] ) << 8 | regGyro[2];
+    gyro_z_raw = ( (int16_t) regGyro[5] ) << 8 | regGyro[4];
 #endif
 
 /* Export Sensor Readouts */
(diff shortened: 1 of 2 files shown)


imu/imu.c
```diff
@@ -417,9 +417,9 @@ if ( imu_status != IMU_OK )
     gyro_y_raw = ( (uint16_t) regGyro[0] ) << 8 | regGyro[1];
     gyro_z_raw = ( (uint16_t) regGyro[0] ) << 8 | regGyro[1];
 #elif defined( A0002_REV2 )
-    gyro_x_raw = ( (uint16_t) regGyro[1] ) << 8 | regGyro[0];
-    gyro_y_raw = ( (uint16_t) regGyro[3] ) << 8 | regGyro[2];
-    gyro_z_raw = ( (uint16_t) regGyro[5] ) << 8 | regGyro[4];
+    gyro_x_raw = ( (int16_t) regGyro[1] ) << 8 | regGyro[0];
```
ETSells: are we sure this is gonna preserve signed data correctly since the msb holds the sign bit?

consider memcpy-ing 6 bytes from the uint regGyro container to the start of the gyro raw structure for safety

imu/imu.c
```diff
@@ -477,12 +477,12 @@ if ( imu_status != IMU_OK )
 	}
 
 /* Combine high byte and low byte to 16 bit data  */
-accel_x_raw = ( (uint16_t) regRaw[1] ) << 8 | regRaw[0];
-accel_y_raw = ( (uint16_t) regRaw[3] ) << 8 | regRaw[2];
-accel_z_raw = ( (uint16_t) regRaw[5] ) << 8 | regRaw[4];
-gyro_x_raw = ( (uint16_t) regRaw[7] ) << 8 | regRaw[6];
-gyro_y_raw = ( (uint16_t) regRaw[9] ) << 8 | regRaw[8];
-gyro_z_raw = ( (uint16_t) regRaw[11] ) << 8 | regRaw[10];
+accel_x_raw = ( (int16_t) regRaw[1] ) << 8 | regRaw[0];
```
ETSells: same as above -- caution sign bit

imu/imu.c
```diff
@@ -1137,12 +1137,12 @@ if( !imu_data_ready && !mag_data_ready )
     Handle IT signal
     ------------------------------------------------------------------------------*/
     /* Combine high byte and low byte to 16 bit data  */
-    imu_raw_processed.accel_x = ( (uint16_t) imu_raw_buffer[1] ) << 8 | imu_raw_buffer[0];
-    imu_raw_processed.accel_y = ( (uint16_t) imu_raw_buffer[3] ) << 8 | imu_raw_buffer[2];
-    imu_raw_processed.accel_z = ( (uint16_t) imu_raw_buffer[5] ) << 8 | imu_raw_buffer[4];
-    imu_raw_processed.gyro_x = ( (uint16_t) imu_raw_buffer[7] ) << 8 | imu_raw_buffer[6];
-    imu_raw_processed.gyro_y = ( (uint16_t) imu_raw_buffer[9] ) << 8 | imu_raw_buffer[8];
-    imu_raw_processed.gyro_z = ( (uint16_t) imu_raw_buffer[11] ) << 8 | imu_raw_buffer[10];
+    imu_raw_processed.accel_x = ( (int16_t) imu_raw_buffer[1] ) << 8 | imu_raw_buffer[0];
```
ETSells: same as above -- caution sign bit

ETSells (commented):
Not sure if I can safely approve bc I'm not sure if the sign is preserved. Consider testing on the emulator since we simulate accel/gyro values.

NArmistead:
Yeah I did the cast wrong. I should be able to cast it to signed outside all the bit logic because that will just reinterpret the MSB as the sign bit

ETSells:
hmmmm... the behavior when casting unsigned to signed above the max sint value is implementation defined. it'll do what you want on gcc but its not in the c standard

NArmistead:
Would we prefer a memcpy then?

ETSells:
I would, for readability and because its predictable. If your casts work then its not an issue tho

NArmistead:
We would need to add intermediate values for imu_it_handler to do the memcpy though

ETSells:
these should be contiguous in both data structures, right?

NArmistead:
They are, so I guess I could make some buffer that flips the raw buffer to little endian then memcpy the whole thing

ETSells:
OHHH its an endianness thing? gross. in that case, consider htons() or a similar function to do the endianness flip. but then again, yeah theres an intermediate buffer. okay, you're right that casting is the cleanest solution if we can do it reliably x-platform.

NArmistead:
From what I can find, GCC, Clang, and MSVC will all reinterpret the bits so it should work. Just to be sure, I'm still gonna try to double check on the emulator

NArmistead:
This is still pretty ugly unfortunately but it gets the job done

ETSells (approved):
Yea this still looks fine.

PR #36, opened 2026-06-06
Change I2C sensor flags to true atomics

ETSells:
## Description
Currently, we use boolean flags to signal when sensor data is ready. These are not thread-safe, and a read->modify->write operation can be interrupted, possibly resulting in undefined behavior in very **VERY** rare cases. This PR eliminates that possibility by using the stdatomic atomic_bool.

Tagging @butchhartman and @NArmistead both on this one since I think @butchhartman might have more experience with thread safety.

### Issue Link
Parent: https://github.com/SunDevilRocketry/Flight-Computer-Firmware/pull/280

### Testing
- [ ] Unit tests modified
- [X] Integration test performed

### Other

### Standards
- [ ] Code is testable

### Error Handling

### Memory

### Performance
- [ ] Busy waiting is avoided

### DIFF
--- a/baro/baro.c
+++ b/baro/baro.c
@@ -26,7 +26,7 @@
 #include <stdbool.h>
 #include <string.h>
 #include <math.h>
-
+#include <stdatomic.h>
 
 /*------------------------------------------------------------------------------
  Project Includes                                                                     
@@ -34,7 +34,6 @@
 #include "main.h"
 #include "sdr_pin_defines_A0002.h"
 #include "baro.h"
-
 
 /*------------------------------------------------------------------------------
 Global Variables  
@@ -49,7 +48,7 @@ static BARO_CAL_DATA baro_cal_data;
 uint8_t baro_raw_buffer[6];
 float baro_pres_proc = NAN;
 float baro_temp_proc = NAN;
-static bool baro_data_ready = false;
+static atomic_bool baro_data_ready = false;
 
 
 /*------------------------------------------------------------------------------
--- a/imu/imu.c
+++ b/imu/imu.c
@@ -24,6 +24,7 @@
  Standard Includes                                                              
 ------------------------------------------------------------------------------*/
 #include <string.h>
+#include <stdatomic.h>
 
 /*------------------------------------------------------------------------------
  Project Includes                                                               
@@ -52,10 +53,10 @@
 #if defined( A0002_REV2 )
 uint8_t imu_raw_buffer[12];
 IMU_RAW imu_raw_processed;
-static bool imu_data_ready;
+static atomic_bool imu_data_ready;
 
 uint8_t mag_raw_buffer[8];
-static bool mag_data_ready;
+static atomic_bool mag_data_ready;
 #endif
 
 MAG_TRIM mag_trim;


baro/baro.c
```diff
@@ -49,7 +50,7 @@ static BARO_CAL_DATA baro_cal_data;
 uint8_t baro_raw_buffer[6];
 float baro_pres_proc = NAN;
 float baro_temp_proc = NAN;
-static bool baro_data_ready = false;
+static atomic_bool baro_data_ready = ATOMIC_VAR_INIT(false);
```
butchhartman: CR: `ATOMIC_VAR_INIT` is deprecated and not needed. Just initialize to false.

baro/baro.c
```diff
@@ -877,7 +878,14 @@ bool baro_get_baro_data_ready
...
 {
-return baro_data_ready;
+bool is_ready = false;
+
+if ( atomic_load(&baro_data_ready) )
+    {
+    is_ready = true;
+    }
+
+return is_ready;
```
butchhartman: Why not directly return value from `atomic_load`?

imu/imu.c
```diff
@@ -52,10 +53,10 @@
 #if defined( A0002_REV2 )
 uint8_t imu_raw_buffer[12];
 IMU_RAW imu_raw_processed;
-static bool imu_data_ready;
+static atomic_bool imu_data_ready = ATOMIC_VAR_INIT(false);
 
 uint8_t mag_raw_buffer[8];
-static bool mag_data_ready;
+static atomic_bool mag_data_ready = ATOMIC_VAR_INIT(false);
```
butchhartman: See previous comment about `ATOMIC_VAR_INIT`
(line 56 as well)

imu/imu.c
```diff
@@ -885,7 +886,14 @@ bool imu_get_imu_data_ready
...
 {
-return imu_data_ready;
+bool is_ready = false;
+
+if ( atomic_load(&imu_data_ready) )
+    {
+    is_ready = true;
+    }
+
+return is_ready;
```
butchhartman: Same comment, why not directly return?
(line 922 as well)

butchhartman (changes requested):
A little CR and some questions on `atomic_load` usage.
Also, consider dropping the usage of `atomic_store` and `atomic_load`. They seem to be unneeded as you are not requesting an explicit memory order. I lowkey think all you had to do was change the type of the flags.

ETSells:
oh damn, thats all it takes?

im unfamiliar with stdatomic unfortunately, and I was treating it as if it was significantly different from a normal bool. the examples I saw showed these functions being used and I didnt realize it'd just work basically as-is.

butchhartman (approved):
Image

butchhartman:
From cppreference:
https://en.cppreference.com/c/language/atomic
https://en.cppreference.com/c/atomic/ATOMIC_VAR_INIT
image
image

...make sure to test it pretty good though because there's always the chance I'm totally wrong lol

ETSells:
Yeah, will do. Will probably re-test the top level once I fix this CI issue on the integration tests

PR #49, opened 2026-07-16
Baro: Initial R3 Driver Implementation

### LINKED ISSUE #8: Baro: Initial R3 Driver Implementation
**Requester: @ETSells **

The IMU and baro are the first two critical sensors we want in place.

We should be able to read 6 contiguous bytes like the I2C interface for temp and pressure. Thus, we might be able to work within the existing baro.c file, your call. We just need both 24-bit values plus processing functions. 

Part Number: MS560702BA03-50

**Additional context**

**Points of Contact**
Add a tag for the component owners, leads, and/or a senior engineer to serve as primary point of contact.

266-750Balloons:
## Description
Implements support for the MS5607-02BA03 barometer on the revision 3 Flight Computer. Not yet finished, currently a draft.

The way a read works in this implementation is it goes through a finite state machine to:
1. Initiate internal baro conversion for pressure.
2. Initiate the timeout required by the datasheet for pressure conversion.
3. Run the ADC read for pressure.
4. Initiate internal baro conversion for temperature.
5. Initiate the timeout required by the datasheet for temperature conversion.
6. Run the ADC read for temperature.
7. Move the FSM to the successful terminal state.

The getter function will then use the conversion factors to convert to useful float values (not yet implemented).

### Issue Link
#8
 
Also see corresponding firmware PR, SunDevilRocketry/Flight-Computer-Firmware#303, which has examples of the interrupt handlers that will be needed for this.

### Testing
- [ ] Unit tests modified

### Other

### Standards
- [ ] Code is testable

### Error Handling

### Memory

### Performance
- [ ] Busy waiting is avoided

### DIFF
--- a/baro/baro.c
+++ b/baro/baro.c
@@ -950,7 +950,7 @@ else
 *                                                                              *
 * DESCRIPTION:                                                                 * 
 * 		ISR for baro data reception. Heavy ISR.	Consider moving processing to  *
-*		baro_get_it.								                           *
+*		baro_get_IT.								                           *
 *                                                                              *
 *******************************************************************************/
 BARO_STATUS baro_IT_handler
@@ -1003,13 +1003,13 @@ return baro_status;
 /*******************************************************************************
 *                                                                              *
 * PROCEDURE:                                                                   * 
-* 		baro_get_it                   		                               	   *
+* 		baro_get_IT                   		                               	   *
 *                                                                              *
 * DESCRIPTION:                                                                 * 
 * 		Getter function for IT baro data.				                       *
 *                                                                              *
 *******************************************************************************/
-BARO_STATUS baro_get_it
+BARO_STATUS baro_get_IT
 	(
 	float* pres_ptr, /* o: pressure */
 	float* temp_ptr  /* o: temperature */
@@ -1024,7 +1024,7 @@ if( !baro_data_ready )
 *temp_ptr = baro_temp_proc;
 
 return BARO_OK;
-} /* baro_get_it */
+} /* baro_get_IT */
 
 
 /*******************************************************************************
--- a/baro/baro.c
+++ b/baro/baro.c
@@ -1,27 +1,27 @@
 /*******************************************************************************
 *
-* FILE: 
+* FILE:
 * 		baro.c
 *
-* DESCRIPTION: 
+* DESCRIPTION:
 * 		Contains API functions for the barometric pressure sensor
 *
-* COPYRIGHT:                                                                   
-*       Copyright (c) 2025 Sun Devil Rocketry.                                 
-*       All rights reserved.                                                   
-*                                                                              
-*       This software is licensed under terms that can be found in the LICENSE 
-*       file in the root directory of this software component.                 
-*       If no LICENSE file comes with this software, it is covered under the   
-*       BSD-3-Clause.                                                          
-*                                                                              
-*       https://opensource.org/license/bsd-3-clause          
+* COPYRIGHT:
+*       Copyright (c) 2025 Sun Devil Rocketry.
+*       All rights reserved.
+*
+*       This software is licensed under terms that can be found in the LICENSE
+*       file in the root directory of this software component.
+*       If no LICENSE file comes with this software, it is covered under the
+*       BSD-3-Clause.
+*
+*       https://opensource.org/license/bsd-3-clause
 *
 *******************************************************************************/
 
 
 /*------------------------------------------------------------------------------
- Standard Includes                                                                     
+ Standard Includes
 ------------------------------------------------------------------------------*/
 #include <stdbool.h>
 #include <string.h>
@@ -29,14 +29,14 @@
 #include <stdatomic.h>
 
 /*------------------------------------------------------------------------------
- Project Includes                                                                     
+ Project Includes
 ------------------------------------------------------------------------------*/
 #include "main.h"
 #include "sdr_pin_defines_A0002.h"
 #include "baro.h"
 
 /*------------------------------------------------------------------------------
-Global Variables  
+Global Variables
 ------------------------------------------------------------------------------*/
 
 /* Current Baro Sensor configuration */
@@ -52,7 +52,7 @@ static atomic_bool baro_data_ready = false;
 
 
 /*------------------------------------------------------------------------------
- Internal function prototypes 
+ Internal function prototypes
 ------------------------------------------------------------------------------*/
 
 /* Converts two bytes to uint16_t format */
@@ -85,6 +85,12 @@ static BARO_STATUS write_reg
 	uint8_t  data      /* In: Register contents           */
 	);
 
+/* Configure/intialize the barometric pressure sensor */
+static BARO_STATUS baro_config
+		(
+		BARO_CONFIG* config_ptr
+		);
+
 /* Load the compensation data from the baro sensor */
 static BARO_STATUS load_cal_data
 	(
@@ -98,7 +104,7 @@ static float temp_compensate
 	);
 
 /* Apply the compensation formula to raw pressure readouts */
-static float press_compensate 
+static float press_compensate
 	(
 	uint32_t raw_readout
 	);
@@ -117,16 +123,16 @@ static BARO_STATUS baro_flush_fifo
 
 
 /*------------------------------------------------------------------------------
- API Functions 
+ API Functions
 ------------------------------------------------------------------------------*/
 
 
 /*******************************************************************************
 *                                                                              *
-* PROCEDURE:                                                                   * 
+* PROCEDURE:                                                                   *
 *       baro_init                                                              *
 *                                                                              *
-* DESCRIPTION:                                                                 * 
+* DESCRIPTION:                                                                 *
 * 		Intialize the barometric pressure sensor                               *
 *                                                                              *
 *******************************************************************************/
@@ -136,7 +142,7 @@ BARO_STATUS baro_init
 	)
 {
 /*------------------------------------------------------------------------------
- Local variables  
+ Local variables
 ------------------------------------------------------------------------------*/
 BARO_STATUS       baro_status;    /* Status code from Baro API calls   */
 uint8_t           baro_device_id; /* Baro device id                    */
@@ -145,7 +151,7 @@ float             init_temp;      /* Temperature read after init       */
 
 
 /*------------------------------------------------------------------------------
- Initializations 
+ Initializations
 ------------------------------------------------------------------------------*/
 baro_status    = BARO_OK;
 baro_device_id = 0;
@@ -154,7 +160,7 @@ init_temp      = 0;
 
 
 /*------------------------------------------------------------------------------
- API Function Implementation 
+ API Function Implementation
 ------------------------------------------------------------------------------*/
 
 /* Disable IRQ */
@@ -181,8 +187,8 @@ if ( baro_status != BARO_OK )
 HAL_Delay( 10 );
 
 /* Check the Baro error register */
-baro_status = read_regs( BARO_REG_ERR_REG      , 
-                         sizeof( baro_err_reg ), 
+baro_status = read_regs( BARO_REG_ERR_REG      ,
+                         sizeof( baro_err_reg ),
 						 &baro_err_reg );
 if ( baro_status != BARO_OK )
 	{
@@ -210,7 +216,7 @@ if ( baro_status != BARO_OK )
 /* Initialize the compensation temperature */
 baro_status = baro_get_temp( &init_temp );
 if ( baro_status != BARO_OK )
-	{ 
+	{
 	return baro_status;
 	}
 
@@ -229,20 +235,20 @@ return baro_status;
 
 /*******************************************************************************
 *                                                                              *
-* PROCEDURE:                                                                   * 
+* PROCEDURE:                                                                   *
 *       baro_config                                                            *
 *                                                                              *
-* DESCRIPTION:                                                                 * 
+* DESCRIPTION:                                                                 *
 * 		Configure the barometric pressure sensor                               *
 *                                                                              *
 *******************************************************************************/
-BARO_STATUS baro_config
+static BARO_STATUS baro_config
 	(
 	BARO_CONFIG* config_ptr
 	)
 {
 /*------------------------------------------------------------------------------
- Local variables  
+ Local variables
 ------------------------------------------------------------------------------*/
 BARO_STATUS   baro_status; /* Baro API call return codes                  */
 uint8_t       pwr_ctrl;    /* Contents of PWR_CTRL register               */
@@ -252,7 +258,7 @@ uint8_t       iir;         /* Contents of CONFIG register (IIR Filter)    */
 
 
 /*------------------------------------------------------------------------------
- Initializations 
+ Initializations
 ------------------------------------------------------------------------------*/
 baro_status = BARO_OK;
 pwr_ctrl    = 0;
@@ -262,7 +268,7 @@ iir         = 0;
 
 
 /*------------------------------------------------------------------------------
- Extract Baro settings from config struct  
+ Extract Baro settings from config struct
 ------------------------------------------------------------------------------*/
 
 /* Set register contents */
@@ -283,7 +289,7 @@ baro_configuration.IIR_setting       = config_ptr -> IIR_setting;
 
 
 /*------------------------------------------------------------------------------
- API function implementation 
+ API function implementation
 ------------------------------------------------------------------------------*/
 
 /* Write to the PWR_CTRL register -> Operating mode and enable sensors */
@@ -322,32 +328,32 @@ return BARO_OK;
 
 /*******************************************************************************
 *                                                                              *
-* PROCEDURE:                                                                   * 
+* PROCEDURE:                                                                   *
 * 		baro_get_device_id                                                     *
 *                                                                              *
-* DESCRIPTION:                                                                 * 
+* DESCRIPTION:                                                                 *
 * 		Gets the device ID of the barometric pressure sensor, primarily used   *
 *       to verify that the sensor can be accessed by the MCU                   *
 *                                                                              *
 *******************************************************************************/
 BARO_STATUS baro_get_device_id
 	(
-   	uint8_t* baro_id_ptr /* Out: Baro device id */ 
+   	uint8_t* baro_id_ptr /* Out: Baro device id */
 	)
 {
 /*------------------------------------------------------------------------------
- Local Variables 
+ Local Variables
 ------------------------------------------------------------------------------*/
 BARO_STATUS baro_status; /* Status codes returned from baro API calls */
 
 
 /*------------------------------------------------------------------------------
- API Function implementation 
+ API Function implementation
 ------------------------------------------------------------------------------*/
 
 /* Read baro register with I2C */
-baro_status = read_regs( BARO_REG_CHIP_ID, 
-                         sizeof( uint8_t ), 
+baro_status = read_regs( BARO_REG_CHIP_ID,
+                         sizeof( uint8_t ),
 						 baro_id_ptr );
 return baro_status;
 } /* baro_get_device_id */
@@ -355,10 +361,10 @@ return baro_status;
 
 /*******************************************************************************
 *                                                                              *
-* PROCEDURE:                                                                   * 
+* PROCEDURE:                                                                   *
 * 		baro_get_pressure                                                      *
 *                                                                              *
-* DESCRIPTION:                                                                 * 
+* DESCRIPTION:                                                                 *
 * 		retrieves a pressure reading from the sensor                           *
 *                                                                              *
 *******************************************************************************/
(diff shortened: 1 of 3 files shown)


baro/baro.c
ETSells, line 1: as far as i can tell, this is almost exclusively whitespace changes

ETSells, line 492: has this always been a stub? need to fix

baro/baro.h
ETSells, line 1: you might have something like clang-format enabled. please ensure these whitespace changes are not present in the final PR, as they make diffs very hard to read.

baro/baro.h
```diff
@@ -112,36 +123,59 @@ typedef enum _BARO_SENSOR_ENABLES
...
 	BARO_PRESS_OSR_X1 = 0,
     BARO_PRESS_OSR_X2    ,
 	BARO_PRESS_OSR_X4    ,
 	BARO_PRESS_OSR_X8    ,
 	BARO_PRESS_OSR_X16   ,
 	BARO_PRESS_OSR_X32
+	#endif
+	/* Prevent any possibility that vals get misinterpreted between revs */
+	#ifdef A0010
+	BARO_PRESS_OSR_X256 = 6,
```
ETSells, line 168: are these really the only settings for rev 3? id assume much lower oversampling rates are possible

baro/baro.h
```diff
@@ -166,32 +200,42 @@ typedef enum _BARO_IIR_SETTING
...
 	} BARO_IIR_SETTING;
+#endif
 
 /* Baro sensor configuration settings struct */
 typedef struct _BARO_CONFIG
 	{
 	/* Sensor enables */
 	BARO_SENSOR_ENABLES enable;
 
+	#ifdef A0002_REV2
```
ETSells: nit: we should place platform dependent options in a consistent place in memory (e.g. common items, then rev 2 items conddef'd, then rev 3)

baro/baro.h
```diff
@@ -166,32 +200,42 @@ typedef enum _BARO_IIR_SETTING
...
 	BARO_ODR_SETTING ODR_setting;
 
 	/* IIR Filter Coefficient Selection */
 	BARO_IIR_SETTING IIR_setting;
-
+	#endif
 	} BARO_CONFIG;
 
+/* These structs are static to the rev3 implementation.
+ * They should be for rev2 as well, but aren't.
```
ETSells: good justification.

we could consider stronger encapsulation via a private header and/or renaming these structs to follow our private scoping convention (preceded by underscores like python)

baro/baro.h
```diff
@@ -239,53 +283,60 @@ typedef struct _BARO_CAL_DATA
...
+BARO_STATUS baro_get_device_id
 	(
-	BARO_CONFIG* config_ptr 
+   	uint8_t* baro_id
 	);
+#endif
 
+#ifdef A0010
 /* verifies sensor can be accessed */
+/* The serial is 12 bits on the rev 3 sensor */
```
ETSells, line 327: tbh these don't need to be publically accessible APIs right? we could make these static for each so there's less platform dependent switching.

baro/baro.h
```diff
@@ -239,53 +283,60 @@ typedef struct _BARO_CAL_DATA
...
 	);
 
 /* gets temp data from sensor */
 BARO_STATUS baro_get_temp
 	(
-    float* temp_ptr 
+    float* temp_ptr
 	);
 
 /* converts pressure and temp data into altitude --> do research on formula */
```
ETSells, line 349: this function should be removed. our ISA model is higher level sensor data processing and lives in sensor.c

this is not scoped to your changes, but it's an observation about the life cycle data under review and should be considered as you do the rest of this.

baro/baro.h
```diff
@@ -239,53 +283,60 @@ typedef struct _BARO_CAL_DATA
...
+#endif
 
 /* returns the baro_data_ready flag */
 bool baro_get_baro_data_ready
     (
-    void 
+    void
     );
 
 BARO_STATUS start_baro_read_IT();
```
ETSells: while we're here, can we consider renaming these functions? i've been informally starting a convention where public APIs of a module are preceded by that module's name, so it'd be:

baro_start_read_it()
baro_it_handler()
baro_get_it() (or something more accurate, like baro_get_latest())

the idea for this is to allow intellisense to narrow down the relevant function name by only presenting functions for a given module by the time you've typed that function name, so typing `baro_` would give you the above as suggestions. i've seen it help a bit in larger codebases.

statics are not required to follow this convention as they're only accessible within that file. it's encouraged for private functions (scoped to a module but not just one file) to follow the convention to make it easier for that module to be maintained.

baro/baro_r3.c
```diff
@@ -0,0 +1,331 @@
...
+#include "stm32h7xx_hal.h"
+#include "sdr_pin_defines_A0010.h"
+#include "baro.h"
+
+/* Type Definitions ----------------------------------------------------------*/
+
+/* Baro calibration data struct in integer format */
+typedef struct _BARO_CAL_DATA_INT
+	{
+	uint16_t par_c1; // Pressure Sensitity
```
ETSells, line 38: nit: // comments are typically reserved for temp comments so it's easier to find them later. even for single line comments, we prefer the multiline `/* comment */` format.

not sure if this was enforced that much by colton, but i've noticed that the `//` comments are rare so i'd like to keep it.

baro/baro_r3.c
```diff
@@ -0,0 +1,331 @@
...
+ */
+static uint32_t pressure_timeout;
+static uint32_t temperature_timeout;
+
+/* Raw SPI bytes with readings */
+uint8_t baro_raw_temp_buffer[3];
+uint8_t baro_raw_press_buffer[3];
+
+/* State of barometer read process */
+static _Atomic BARO_READ_STATE baro_read_state = BARO_READ_FAIL;
```
ETSells, line 84: note: this is one of those things that would really benefit from C23's typed enums. we should consider accelerating our switch to c23 so we can make sure things like atomics/shared variables have the expected underlying type.

cc: @NArmistead

baro/baro_r3.c
```diff
@@ -0,0 +1,331 @@
...
+
+static BARO_STATUS success;
```
ETSells, line 86: nit: let's give these statics a default state before init.

baro/baro_r3.c
```diff
@@ -0,0 +1,331 @@
...
+/* Private function prototypes -----------------------------------------------*/
+
+/* Some of these are probably good candidates for macros */
+
+/*
+ * Implementation Notes
+ * I think we can create a TIM_OC_InitTypeDef, configure HAL_TIM_OC_ConfigChannel in a channel,
+ * and then run HAL_TIM_OC_Start. Our callback is then HAL_TIM_OC_DelayElapsedCallback().
+ */
+static BARO_STATUS create_timer_interrupt // TODO actually implement
```
ETSells: consider using @NArmistead's task scheduler instead of reserving a hardware timer for a single driver.

baro/baro_r3.c
```diff
@@ -0,0 +1,331 @@
...
+{
+switch(baro_read_state) {
+    case BARO_READ_FAIL:
+    case BARO_READ_DONE: // No currently running read
+        baro_read_state = BARO_CONV_PRESSURE;
+        baro_IT_handler(); // Keep FSM logic in that function
+        break;
+    default: // Read in progress
+        return BARO_BUSY;
+}
```
ETSells: nit: i know this is a draft, but this is already hard to read due to divergence from SDR's coding standards

baro/baro_r3.c
```diff
@@ -0,0 +1,331 @@
...
+BARO_STATUS start_baro_read_IT
+    (
+    void
+    )
+{
+switch(baro_read_state) {
+    case BARO_READ_FAIL:
+    case BARO_READ_DONE: // No currently running read
+        baro_read_state = BARO_CONV_PRESSURE;
+        baro_IT_handler(); // Keep FSM logic in that function
```
ETSells: cr: Do not call an ISR from a blocking context. Ideally, you abstract the FSM logic out to a shared function, and then this start function & the ISR can call into it.

baro/baro_r3.c
```diff
@@ -0,0 +1,331 @@
...
+  * @brief Interrupt service routine for data acquisition
+  *
+  * @details This function must be called by three interrupt callbacks:
+  * SPI_TxCpltCallback(), HAL_SPI_TxRxCpltCallback(), and
+  * HAL_TIM_OC_DelayElapsedCallback(). These bring the read through a series of
+  * 7 states (plus a done and fail state), needed to account for timeouts required by the chip.
+  *
+  * @retval Status of the barometer
+  */
+BARO_STATUS baro_IT_handler
```
ETSells, line 390: An ISR will never be able to report its own status back to the main thread via return. You'll need to do this via a static.

baro/baro_r3.c
```diff
@@ -0,0 +1,331 @@
...
+switch(baro_read_state) {
+    // TODO actually deal with chip select pin
+    case BARO_CONV_PRESSURE:
+        // 1. Called by start_baro_read_IT, start pressure conversion
+        success = transmit_cmd_IT(pressure_cmd);
+        update_state(BARO_WAIT_PRESSURE);
+        break;
+    case BARO_WAIT_PRESSURE:
+        // 2. Called by SPI_TxCpltCallback(), create pressure conversion timeout
+        success = create_timer_interrupt(pressure_timeout);
```
ETSells, line 414: how long is this delay? on second thought, this might actually need microsecond precision which means the scheduler wouldn't work for it

baro/baro_r3.c
```diff
@@ -0,0 +1,331 @@
...
+        success = transceive_adc_IT(baro_raw_press_buffer);
+        update_state(BARO_READ_FINISH);
+        break;
+    case BARO_READ_FINISH:
+        // 7. Called by HAL_SPI_TxRxCpltCallback(), switch to done state.
+        success = BARO_OK;
+        update_state(BARO_READ_DONE);
+        break;
+    case BARO_READ_DONE:
+        // We're done. Nothing happens.
```
ETSells, line 470: this state should not be entered during normal execution. consider debug asserting if it is hit.

ETSells (commented):
thoughts on the first cut

baro/baro.h
266-750Balloons: I wouldn't be against that.

baro/baro.c
266-750Balloons: Oops. Will fix that. It's probably because I switched from VSCode to Zed and some of my configs are out of wack.

266-750Balloons: As far as I can tell.

baro/baro.h
266-750Balloons: Will do.

baro/baro.h
266-750Balloons: As far as I can tell. At the very least, there are all that's mentioned in the datasheet.

baro/baro_r3.c
266-750Balloons: Would this be better:
```C
switch ( baro_read_state )
    {
    case BARO_READ_FAIL:

    case BARO_READ_DONE: // No currently running read
        baro_read_state = BARO_CONV_PRESSURE;
        baro_IT_handler(); // Keep FSM logic in that function
        break;

    default: // Read in progress
        return BARO_BUSY;
    }
}
```
If not, what in particular is the issue?

baro/baro_r3.c
266-750Balloons: Do you mean don't call the ISR from a blocking context?

baro/baro_r3.c
266-750Balloons: I don't think this is actually used for anything. It's mostly because that's what was already in the header file and the original implementation. Should we just make the handler have a void return type, in both the original and new implementations?

baro/baro_r3.c
266-750Balloons: Alright. Will do.

baro/baro_r3.c
266-750Balloons: pressure_timeout and temperature_timeout will be set by baro_init based on the OSR (as I don't want to calculate them at runtime); the delay depends on the sample rate. Here's the table from the datasheet.
image
I think I've found a way to get the precision we need. You can just do an output capture interrupt using the channels feature of our existing microsecond timer. It has 4 channels for output capture, and I don't think we're using them if I understand correctly.

baro/baro_r3.c
ETSells: Basically what I'm trying to do here is make sure anything that calls itself an "interrupt handler" is safe from race conditions arising from lack of consideration of re-entrancy. If it says "*_IT_handler" I expect it'll only be called from the context of an ISR from the NVIC that can only be triggered once at a time. If we enter this from a blocking context and then the ISR triggers, we could be in the middle of modifying state and end up with a really nasty to debug race condition.

baro/baro_r3.c
ETSells: oh god this really isn't that bad, i was just way too tired to be doing a review at the time, sorry drew.

yes, the suggested version is good, although having FAIL fallthrough to DONE feels like it'll be an issue.

baro/baro.c
266-750Balloons: Note to self: commit is for baro_get_altitude.

The context is less clear because I finally fixed the white space f*** up.

ETSells: We definitely want to get rid of it bc the actual conversion is done by sensor.

266-750Balloons:
This obviously isn't complete (there's certainly some integer logic that I'll need to examine when we get the actual rev3), but I figure this should be in the main tree sooner than later.

baro/baro.h
```diff
@@ -95,9 +100,18 @@ typedef enum _BARO_STATUS
 	BARO_UNRECOGNIZED_HAL_STATUS,
 	BARO_UNSUPPORTED_CONFIG     ,
 	BARO_UNRECOGNIZED_CHIP_ID   ,
+	#ifdef A00100               ,
```
ETSells: macro is not defined on rev 3 -- fix to A0010

ETSells: can we do the cond-defs at the very end here? right now there's 3 blocks, it could be consolidated to two

baro/baro.h
ETSells: still true. little point in having everything live in the same header if we'll need platform dependent handling for every public API anyway

ETSells:
@266-750Balloons please also open the mod PR, looks like you have a branch on that submod

ETSells (changes requested):
Significant build issues present on R3, so I've halted this review. Please add your new driver to the rev3 build system and re-request review.

ref:
```
$ make -j4
arm-none-eabi-gcc -c -mcpu=cortex-m7 -mthumb -mfpu=fpv5-d16 -mfloat-abi=hard -DUSE_PWR_LDO_SUPPLY -DUSE_USB_CDC_FS -DUSE_HAL_DRIVER -DA0010 -DSTM32H733xx -I. -I../../mod/error_sdr -I../../mod/math_sdr -I../../driver/baro -I../../driver/led -I../../driver/lora -I../../driver/flash -I../../driver/usb -I../../driver/usb/USB_DEVICE/App -I../../driver/usb/USB_DEVICE/Target -I../../driver/gps -I../../driver/buzzer -I../../init/rev3/config/Inc -I../../lib -I../../lib/Drivers/STM32H7xx_HAL_Driver/Inc -I../../lib/Drivers/STM32H7xx_HAL_Driver/Inc/Legacy -I../../lib/Middlewares/ST/STM32_USB_Device_Library/Core/Inc -I../../lib/Middlewares/ST/STM32_USB_Device_Library/Class/CDC/Inc -I../../lib/Drivers/CMSIS/Device/ST/STM32H7xx/Include -I../../lib/Drivers/CMSIS/Include -Og -Wall -fdata-sections -ffunction-sections -g -gdwarf-2 -MMD -MP -MF"build/main.d" -Wa,-a,-ad,-alms=build/main.lst main.c -o build/main.o
arm-none-eabi-gcc -c -mcpu=cortex-m7 -mthumb -mfpu=fpv5-d16 -mfloat-abi=hard -DUSE_PWR_LDO_SUPPLY -DUSE_USB_CDC_FS -DUSE_HAL_DRIVER -DA0010 -DSTM32H733xx -I. -I../../mod/error_sdr -I../../mod/math_sdr -I../../driver/baro -I../../driver/led -I../../driver/lora -I../../driver/flash -I../../driver/usb -I../../driver/usb/USB_DEVICE/App -I../../driver/usb/USB_DEVICE/Target -I../../driver/gps -I../../driver/buzzer -I../../init/rev3/config/Inc -I../../lib -I../../lib/Drivers/STM32H7xx_HAL_Driver/Inc -I../../lib/Drivers/STM32H7xx_HAL_Driver/Inc/Legacy -I../../lib/Middlewares/ST/STM32_USB_Device_Library/Core/Inc -I../../lib/Middlewares/ST/STM32_USB_Device_Library/Class/CDC/Inc -I../../lib/Drivers/CMSIS/Device/ST/STM32H7xx/Include -I../../lib/Drivers/CMSIS/Include -Og -Wall -fdata-sections -ffunction-sections -g -gdwarf-2 -MMD -MP -MF"build/stm32h7xx_it.d" -Wa,-a,-ad,-alms=build/stm32h7xx_it.lst ../../init/rev3/config/Src/stm32h7xx_it.c -o build/stm32h7xx_it.o
arm-none-eabi-gcc -c -mcpu=cortex-m7 -mthumb -mfpu=fpv5-d16 -mfloat-abi=hard -DUSE_PWR_LDO_SUPPLY -DUSE_USB_CDC_FS -DUSE_HAL_DRIVER -DA0010 -DSTM32H733xx -I. -I../../mod/error_sdr -I../../mod/math_sdr -I../../driver/baro -I../../driver/led -I../../driver/lora -I../../driver/flash -I../../driver/usb -I../../driver/usb/USB_DEVICE/App -I../../driver/usb/USB_DEVICE/Target -I../../driver/gps -I../../driver/buzzer -I../../init/rev3/config/Inc -I../../lib -I../../lib/Drivers/STM32H7xx_HAL_Driver/Inc -I../../lib/Drivers/STM32H7xx_HAL_Driver/Inc/Legacy -I../../lib/Middlewares/ST/STM32_USB_Device_Library/Core/Inc -I../../lib/Middlewares/ST/STM32_USB_Device_Library/Class/CDC/Inc -I../../lib/Drivers/CMSIS/Device/ST/STM32H7xx/Include -I../../lib/Drivers/CMSIS/Include -Og -Wall -fdata-sections -ffunction-sections -g -gdwarf-2 -MMD -MP -MF"build/stm32h7xx_hal_msp.d" -Wa,-a,-ad,-alms=build/stm32h7xx_hal_msp.lst ../../init/rev3/config/Src/stm32h7xx_hal_msp.c -o build/stm32h7xx_hal_msp.o
arm-none-eabi-gcc -c -mcpu=cortex-m7 -mthumb -mfpu=fpv5-d16 -mfloat-abi=hard -DUSE_PWR_LDO_SUPPLY -DUSE_USB_CDC_FS -DUSE_HAL_DRIVER -DA0010 -DSTM32H733xx -I. -I../../mod/error_sdr -I../../mod/math_sdr -I../../driver/baro -I../../driver/led -I../../driver/lora -I../../driver/flash -I../../driver/usb -I../../driver/usb/USB_DEVICE/App -I../../driver/usb/USB_DEVICE/Target -I../../driver/gps -I../../driver/buzzer -I../../init/rev3/config/Inc -I../../lib -I../../lib/Drivers/STM32H7xx_HAL_Driver/Inc -I../../lib/Drivers/STM32H7xx_HAL_Driver/Inc/Legacy -I../../lib/Middlewares/ST/STM32_USB_Device_Library/Core/Inc -I../../lib/Middlewares/ST/STM32_USB_Device_Library/Class/CDC/Inc -I../../lib/Drivers/CMSIS/Device/ST/STM32H7xx/Include -I../../lib/Drivers/CMSIS/Include -Og -Wall -fdata-sections -ffunction-sections -g -gdwarf-2 -MMD -MP -MF"build/error_sdr.d" -Wa,-a,-ad,-alms=build/error_sdr.lst ../../mod/error_sdr/error_sdr.c -o build/error_sdr.o
In file included from ../../init/rev3/config/Src/stm32h7xx_it.c:23:                                             
../../driver/baro/baro.h:331:9: error: expected ';', ',' or ')' before 'uint16_t'                               
  331 |         uint16_t* baro_id                                                                               
      |         ^~~~~~~~                                                                                        
../../driver/baro/baro.h:357:1: error: unknown type name 'bool'                                                 
  357 | bool baro_get_baro_data_ready                                                                           
      | ^~~~                                                                                                    
../../driver/baro/baro.h:35:1: note: 'bool' is defined in header '<stdbool.h>'; this is probably fixable by adding '#include <stdbool.h>'                                                                                       
   34 | #include "sdr_pin_defines_A0010.h"
  +++ |+#include <stdbool.h>                                                                                    
   35 | #endif
make: *** [Makefile:236: build/stm32h7xx_it.o] Error 1
make: *** Waiting for unfinished jobs....
```

baro/baro.h
266-750Balloons: I think my reason for the three blocks is I2C and SPI error are spiritually very similar errors, and I wanted them to get the same "number" assigned to them in the enum.

266-750Balloons: Fixed. Whoops.

266-750Balloons:
Alright. I've fixed a lot of the glaring build issues. It compiles now.

Sorry, this was really sloppy. Should be better now. I've got a lot on my plate, so I kind of rushed this PR. I've fixed the build issues, so it compiles; obviously functionality verification will have to wait until real HW.

266-750Balloons:
Also, I agree that in retrospect, we probably should have had different headers, but it's a little late for me to do so now.

I did get rid of the get_device_id function for rev3 because there's little point in it when we have the CRC check.

PR #50, opened 2026-07-19
Fix BMM150 blocking magnetometer read

bjornhbengtsson:
## Description
Fixes the blocking BMM150 magnetometer read path so it is consistent with the interrupt-driven read path.

Changes:
- Read all 8 BMM150 data bytes on A0002 Rev 2, including RHALL
- Replace logical `&&` operations with bitwise `&` masks
- Sign-extend the packed 13-bit X/Y and 15-bit Z measurements
- Store the RHALL measurement in `IMU_RAW`
- Propagate all magnetometer read errors instead of checking only for timeout
- Preserve the six-byte read behavior for A0002 Rev 1

### Issue Link
Related to https://github.com/SunDevilRocketry/Flight-Computer-Firmware/issues/300

### Testing

- [x] Passes existing unit tests
- [x] Unit tests modified
- [x] Integration test performed

- `git diff --check` passed
- No existing host-side unit-test target compiles `driver/imu/imu.c`
- Hardware validation is still required

### Other

This PR fixes the blocking `imu_get_mag_xyz()` function, which is not currently called by the flight firmware, but served as a useful first step in familiarizing myself with the magnetometer implementation. Further work for issue #300 will investigate and validate the interrupt-driven magnetometer path used during normal operation and determine how the retrieved magnetometer data can be used for sensor fusion.

https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bmm150-ds001.pdf

https://getyarn.io/yarn-clip/23a5374e-cb47-431e-82a0-7a63da68f11f/gif

### Standards
- [x] Code is testable

### Error Handling

### Memory

### Performance
- [x] Busy waiting is avoided

### DIFF
--- a/imu/imu.c
+++ b/imu/imu.c
@@ -516,11 +516,12 @@ IMU_STATUS imu_get_mag_xyz
 /*------------------------------------------------------------------------------
  Local variables 
 ------------------------------------------------------------------------------*/
-uint8_t     regMag[6];    /* Magnetometer register bytes      */
-uint16_t    mag_x_raw;    /* Raw magnetometer sensor readouts */ 
-uint16_t    mag_y_raw; 
-uint16_t    mag_z_raw; 
-IMU_STATUS  imu_status;   /* IMU status return codes          */
+uint8_t    regMag[8];        /* Magnetometer register bytes      */
+int16_t    mag_x_raw;        /* Raw magnetometer sensor readouts */
+int16_t    mag_y_raw;
+int16_t    mag_z_raw;
+uint16_t   mag_hall_raw = 0; /* Rev 1 does not provide RHALL     */
+IMU_STATUS imu_status;       /* IMU status return codes           */
 
 
 /*------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


bjornhbengtsson:
The automated GPS test currently fails while compiling `mod/sensor/sensor.h` because `IMU_DATA` is undefined in the test environment. The failure occurs before compiling the modified `imu_get_mag_xyz()` function and appears unrelated to this PR. The servo job was subsequently cancelled.

imu/imu.c
```diff
@@ -516,53 +516,87 @@ IMU_STATUS imu_get_mag_xyz
...
 /* Read MAG_X, MAG_Y, MAG_Z high byte and low byte registers */
-#if   defined( A0002_REV1 )
-    imu_status = read_mag_regs( IMU_REG_MAG_XOUT_H, 
-                                &regMag[0]        , 
-                                sizeof( regMag ) );
-#elif defined( A0002_REV2 )
-    imu_status = read_mag_regs( MAG_REG_DATAX_L, 
-                                &regMag[0]     , 
-                                sizeof( regMag ) );
+#if     defined(A0002_REV1)
```
ETSells: We are dropping support for the rev 1 IMU in #28. Let's consider dropping this path for rev 1 mag reads.

imu/imu.c
```diff
@@ -516,53 +516,87 @@ IMU_STATUS imu_get_mag_xyz
...
-    mag_x_raw  = (   (uint16_t) regMag[1]                         << MAG_XY_MSB_BITSHIFT ) | 
-                 ( ( (uint16_t) regMag[0] && MAG_XY_LSB_BITMASK ) >> MAG_XY_LSB_BITSHIFT );
-    mag_y_raw  = (   (uint16_t) regMag[3]                         << MAG_XY_MSB_BITSHIFT ) | 
-                 ( ( (uint16_t) regMag[2] && MAG_XY_LSB_BITMASK ) >> MAG_XY_LSB_BITSHIFT );
-    mag_z_raw  = (   (uint16_t) regMag[5]                         << MAG_Z_MSB_BITSHIFT  ) | 
-                 ( ( (uint16_t) regMag[4] && MAG_Z_LSB_BITMASK )  >> MAG_Z_LSB_BITSHIFT  );
+#elif defined(A0002_REV2)
+    mag_x_raw =
+        (int16_t)(((uint16_t)regMag[1] << MAG_XY_MSB_BITSHIFT) |
+        (((uint16_t)regMag[0] & MAG_XY_LSB_BITMASK) >>
```
ETSells: OOF no wonder our mag reads on the blocking path always used to be missing precision on the lower bits lmaoooo! Good spot!

imu/imu.c
```diff
@@ -516,53 +516,87 @@ IMU_STATUS imu_get_mag_xyz
...
-#elif defined( A0002_REV2 )
-    mag_x_raw  = (   (uint16_t) regMag[1]                         << MAG_XY_MSB_BITSHIFT ) | 
-                 ( ( (uint16_t) regMag[0] && MAG_XY_LSB_BITMASK ) >> MAG_XY_LSB_BITSHIFT );
-    mag_y_raw  = (   (uint16_t) regMag[3]                         << MAG_XY_MSB_BITSHIFT ) | 
-                 ( ( (uint16_t) regMag[2] && MAG_XY_LSB_BITMASK ) >> MAG_XY_LSB_BITSHIFT );
-    mag_z_raw  = (   (uint16_t) regMag[5]                         << MAG_Z_MSB_BITSHIFT  ) | 
-                 ( ( (uint16_t) regMag[4] && MAG_Z_LSB_BITMASK )  >> MAG_Z_LSB_BITSHIFT  );
+#elif defined(A0002_REV2)
+    mag_x_raw =
+        (int16_t)(((uint16_t)regMag[1] << MAG_XY_MSB_BITSHIFT) |
```
ETSells: nit: this compound is getting a bit hard to read. Consider using the styling that was here before i.e.
```
mag_x_raw = (int16_t)(   (uint16_t) regMag[1]                        << MAG_XY_MSB_BITSHIFT ) |
                     ( ( (uint16_t) regMag[0] & MAG_XY_LSB_BITMASK ) >> MAG_XY_LSB_BITSHIFT );
```
If you're not used to SDR's code styling it might seem trivial or not as readable, but it makes a huge difference to me as a reviewer and to readability throughout our codebase if everything is more uniform. The compound makes it a bit hard to track everything, but the extra spacing and forcing everything to two lines helps tell me how this works:
1. The byte order out of the raw registers is little endian
2. There is a bitmask applied to the least significant bits of each register
3. There is a shift applied to both the MSB and LSB
4. We bitwise OR the MSB and LSB, then store them into an int16_t that houses the scaled raw values.

ETSells (approved):
Picking up this review myself as I am the SME for legacy IMU. Nick, you're still more than welcome to get eyes on it too.

This looks solid! Consider dropping the rev 1 conddefs, but other than that I'm happy with this!

Agreed that there's no worries about the test issue. That happens on the target branch as well, but will be fixed once FCF changes hit main.

bjornhbengtsson:
Thanks for the review Eli! I removed the Rev 1 path from `imu_get_mag_xyz()`. I also reformatted the register unpacking and sign-extension logic to match SDR's style and make the bitmasking and shifting operations easier to follow.

ETSells (approved):
ship it

Image

bjornhbengtsson:
image

